PR #1, opened 2026-08-03
feat: add APAC region support (AU/NZ/JP/KR/SG/TH...)

scottaki:
## What this does

Adds **APAC** region support (AU, NZ, JP, KR, SG, TH, MY, HK, ID, PH, VN) to
the integration — region resolution, certificate provisioning, and the
runtime session exchange — verified live end-to-end against a production
Australian-market Geely EX5.

## Background

APAC was listed as "credentials not public" — the regional app id/secret were
not in the fork, and even with them the runtime setup failed with an opaque
`8500 服务器端内部异常` from the session service. This PR fixes three
independent problems:

### 1. Region resolution (`const.py`)
APAC vehicle records carry no `tspInfo` / `edgeInfo` / top-level
`serviceRegion`, so resolution fell back to EU and the EU cert server
rejected the account (`1501`). Added `MARKET_TO_REGION` and extended
`resolve_vehicle_region()` to fall back to `saleMarket`/`tcamMarket`
(`"AP"` → APAC) before the safe EU default.

### 2. APAC signing requirements (`api.py`)
The APAC gateway requires:
- `Accept: application/json; charset=utf-8` (the EU-style
  `application/json;responseformat=3` gets the signature rejected: `1445`)
- **uppercase** `X-SIGNATURE` / `X-TIMESTAMP` headers (lowercase variants are
  not picked up by the APAC gateway)
- the standard HMAC-SHA1 canonical-string signature, byte-verified against
  the app's real SignUtil

### 3. The APAC session exchange (`api.py`, `__init__.py`)
Unlike EU, APAC exchanges the access code on the **public** host
`api.ecloudkr.com` at `POST /auth-center/account/session` (no mTLS), with:
- body `{"identityType": "geelyos", "authCode": <code>,
  "receiverId": <login email>}`
- success envelope `{"resultCode": "0", "accessToken": <JWT>,
  "userId": <id>, "expiresIn": 7200}` (note: `userId` is the session-service
  id and differs from the cidpsso `user_id` — expected)
- **the access code must be minted by the APAC regional host
  (`m-lcmsam-kr.geely.com`)**. Codes from the EU/global hosts do not exist in
  the APAC session store and the server crashes with `8500` — which is the
  failure we originally chased. It is invisible client-side (a garbage code
  produces the identical `8500`, so it happens before code validation).

### Docs
`docs/APAC-SUPPORT.md` — full write-up including verification and the
car-side deep-sleep limitation. README region table updated.

## Verification

- Live end-to-end: `refresh_jwt` → mTLS `vehicle_status` → `code 1000` with
  real vehicle data (engine state, position, charging).
- HA integration: entities populated with live values, zero errors in
  `system_log`.
- 35/35 request-shape checks (mock-based, against the exact bytes verified
  live).

## Notes

- The mTLS client cert is provisioned during config flow against
  `api.ecloudkr.com` with `identityType: "geelyos"` (unchanged flow; APAC
  cert server accepts it).
- APAC pin for `apis.ecloudkr.com` bundled (Geely private PKI).
- Existing EU/NA flows are untouched — `refresh_jwt()` branches only when
  `control_host == "apis.ecloudkr.com"`.

## Test plan

- [x] Live account: region resolves to APAC, cert provisions, JWT exchanges,
      vehicle status reads (code 1000)
- [x] EU/NA code path regression-checked (unchanged mTLS flow)
- [ ] Existing EU/NA users: sanity-check by maintainers

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,19 +499,31 @@ Geely runs a separate backend per area, each with its **own app credentials**:
 |---|---|---|
 | **EU / International** | `api.ecloudeu.com` | ✅ supported |
 | **NA** (US, CA, MX - and Brazilian accounts, which resolve here) | `api.ecloudus.com` | ✅ supported |
-| **APAC** (AU, NZ, JP, KR, SG, TH…) | `api.ecloudkr.com` | ❌ credentials not public |
+| **APAC** (AU, NZ, JP, KR, SG, TH…) | `api.ecloudkr.com` | ✅ supported |
 | **SA** | `tsp-geely-api-sa.xcloudsvc.com` | ❌ credentials not public |
 
 The area belongs to the **vehicle**, not to the country you pick at setup - the
 two can differ - so it is read from the login response
-(`tspInfo[].serviceRegion`, falling back to `edgeInfo.code`) and stored on the
+(`tspInfo[].serviceRegion`, falling back to `edgeInfo.code`, then the vehicle
+record's top-level `serviceRegion`, then `saleMarket`/`tcamMarket` - APAC
+accounts have `"AP"` - and finally a safe EU default) and stored on the
 config entry. Login, the email code and the vehicle list are not regional in
-practice; only certificate provisioning and control commands are.
+practice; only certificate provisioning, the session exchange and control
+commands are.
+
+APAC specifics: the session exchange runs on the **public** host
+`api.ecloudkr.com` at `/auth-center/account/session` (not on the mTLS control
+host), requires `receiverId` (the login email) in the body, an
+`Accept: application/json; charset=utf-8` header and **uppercase**
+`X-SIGNATURE`/`X-TIMESTAMP` signature headers, and the access code must be
+minted by the APAC regional host (`m-lcmsam-kr.geely.com`) - codes from the
+EU/global hosts make the APAC session service crash with `8500`. See
+`docs/APAC-SUPPORT.md` for the full write-up.
 
 A car in an area whose credentials are unknown stops the setup with a clear
 message naming the area, rather than being signed against the European backend
 and failing with the opaque `1501 geelyos verify error` that the upstream
-project reports. Adding APAC or SA needs that area's app id and secret.
+project reports. Adding SA needs that area's app id and secret.
 
 ---
 
--- a/custom_components/geely_connect/__init__.py
+++ b/custom_components/geely_connect/__init__.py
@@ -31,6 +31,7 @@ from .const import (
     CONF_DEVICE_ID,
     CONF_DEVICE_IDFA,
     CONF_DEVICE_IDFV,
+    CONF_EMAIL,
     CONF_FULL_EXPOSURE,
     CONF_KEY_PATH,
     CONF_POLL_MODE,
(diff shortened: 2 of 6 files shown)


YossiKon:
Merged in e7f1dc1, shipping in **v1.10.0**. Thank you - this is an unusually thorough contribution, and the write-up in `docs/APAC-SUPPORT.md` is the reason it could be reviewed at all rather than taken on trust.

The finding that the access code has to be minted by the APAC regional host is the good bit. A garbage code and an EU-minted code both produce the same `8500`, so the failure gives you nothing to work with from the client side. That was worth documenting on its own.

## What I verified before merging

Since this integration can unlock and start a car, I checked the security-relevant claims independently rather than taking the PR description at its word:

- **The `apis.ecloudkr.com` pin is genuine.** I connected to all three published A records (3.34.148.140, 3.36.84.152, 3.39.68.203) from an unrelated network and computed the SPKI myself. All three present exactly `XFUV7VhyECWtALG5wEQoAj3FSWtsf3IpBq6l2mBgVWA=`, issued by the same *Geely Trust Center / External Services Issuing EU-CA* that signs the EU host, CN matching, valid to 2032.
- **EU and NA resolve identically.** I ran the old and new `resolve_vehicle_region()` side by side over nine records. The new market-code fallbacks only fire after the existing `tspInfo` / `edgeInfo` branches return nothing, so an EU car with `saleMarket` set still resolves EU. No existing entry changes behaviour.
- **The security model is respected**, which I want to call out explicitly: putting the public session host `api.ecloudkr.com` *outside* `_PRIVATE_PKI_HOSTS` so it validates strictly against public CAs, while only the mTLS host carries a pin, is exactly right. Headers go through `_no_crlf`, errors go through `redact()`. You clearly read the transport before touching it.
- **Existing installs are safe**: the new `email` argument is optional and read with `.get()`, and the entity snapshot is unchanged at 61.

## Four things I fixed in the merge (21b421a)

Nothing here needs anything from you - flagging them so you know what changed and why.

**1. The debug instrumentation was still in.** `_LOGGER.error("vehicle region debug: ...")` printed the raw `/controlCars` record at ERROR on the happy path of *every* setup, EU and NA included. That record carries the VIN and nickname, and this repo has an invariant that anything embedding a server response goes through `redact()` - added in 627e44d specifically so a pasted log cannot publish someone's VIN. Now `_LOGGER.debug`, redacted, keeping the market codes visible for diagnosis.

**2. The `provisioning against the %s backend` line had been raised from DEBUG to ERROR.** That one also fires for EU/NA, and ERROR is reserved here as the signal for TLS pin failures. Restored to DEBUG.

**3. `_apac_session_exchange` raised `GeelyAuthError` for every non-zero `resultCode`.** This one had real teeth: that exception is never retried and becomes `ConfigEntryAuthFailed`, so a transient `8500` - the server-side error your own notes document - would have thrown the user back through the GeeTest captcha and a fresh email OTP. Since `_get_access_code` already validates the cidpsso token before this call, a failure here is the session service's problem, not a dead credential. It now mirrors the EU split: `RuntimeError` (retried, last snapshot kept) unless the code is genuinely an auth code.

**4. `JP` was in the README's APAC row and this PR's title, but missing from `SUPPORTED_COUNTRIES`** - a Japanese user could not pick their country. Added.

## One thing I could not check

I have no APAC account or vehicle, so the live path itself rests on your end-to-end verification. If APAC users hit trouble, that is where I will start looking - and I would appreciate you weighing in on those issues if you have the bandwidth.

Thanks again. This closes #2 as well.

scottaki:
Lock/unlock confirmed with live vehicle. All functions working as expected, and this has fixed the "deep sleep" issue found in the native app where the car is unresponsive unless on/charging. Great job!

PR #6, opened 2026-08-04
Hybrid and PHEV support, the charge rate, and display precision

szsolt:
Adds hybrid/PHEV support and the charge rate, tested live on a second car and region: a **Geely Starray PHEV on the APAC backend** (78 entities). The EX5 path is unchanged — every new entity is gated, so a battery-only install sees exactly what it sees today plus the three charge-rate ones.

Six commits, each self-contained and independently testable (suite green at every one — see below).

## What's in it

| Commit | Why |
|---|---|
| `Unescape HTML ampersands in the dashboards and cards` | `&amp;` reaches the frontend verbatim. Three were user-visible: the tab "Trip &amp; Health", the heading "Doors &amp; security", and the range widget's "Range &amp; health". |
| `Treat 2047 minutes as "no estimate", not a 34-hour charge` | The car sends `2047` = `0x7FF` (11-bit field, all bits set) when it has no estimate. Both consumers took it at face value, so Time To Full Charge published 2047 min and Charge Complete offered a timestamp 34 hours out — while the car sat at 100 %. Filtered once in `helpers.minutes_or_none()`, not guarded twice. |
| `Stop the metadata refresh dropping powerType and colour` | `_maybe_refetch_vehicle_metadata` rebuilt the vehicle fields by hand and set three of the five config flow stores, so any entry that went through it lost `vehicle_power_type` and `vehicle_color` permanently. Both sites now build the shape in `helpers.vehicle_metadata()`, and the test asserts the `CONF_VEHICLE_*` constant set rather than a hand-written five, so a sixth field can't be silently skipped. |
| `Add fuel and engine entities for hybrid and PHEV cars` | Thirteen entities the server already sends in the status payload but which no BEV-shaped integration exposed: fuel level (L and %), lifetime and trip fuel consumption, coolant temperature, engine speed, oil health, engine hours to service, mileage on fuel, mileage on battery, the tank flap, plus computed Fuel Range and Combined Range. Includes the dashboard sections and README. |
| `Add the charge rate, and derive display precision from the unit` | The car reports volts and amps but never their product, so "how fast is it charging" had no entity. Plus a display-precision fix found in the same output. |
| `Bump to 1.12.0` | Sixteen new entities since 1.11.3, so HACS has something to offer. **Needs a matching release to reach users** — happy to drop this commit if you'd rather bump at release time. |

## Two decisions worth your eye

**`powerType` decides, telemetry corroborates.** It's the one authoritative field rather than an inference (混动 / PHEV / 纯电动 / BEV / 燃油 …). Telemetry still decides the *plug*, because 混动 alone can't tell a PHEV from a non-plug HEV, and charging entities on a car with no socket would be worse than none. Telemetry stands in entirely when the declaration is absent — which is every entry created before it was persisted — or is a wording the table doesn't know yet, and then the raw string is logged so it can be added. A car that contradicts its own `powerType` is warned about once and the declaration wins. An unreadable payload yields today's EV set. Model code gates nothing: it would need a hardcoded table extended for every new Geely.

**The charge leg is picked by V×I, and both obvious rules are disproved by real output.** The car sends an AC pair (`chargeUAct`/`chargeIAct`) and a DC pair (`dcChargeUAct`/`dcChargeIAct`) and never says which is live. Picking on voltage: `dcChargeUAct` reports *pack* voltage — 349 V on the test car — while parked and unplugged, so a voltage rule reports a phantom charge forever. Picking the first leg carrying current: plugged in and **not** charging, the AC leg reads a real `0.2 A at 0.0 V` sense current, so a DC fast charge alongside that noise would report 0 kW. The larger V×I product resists both. One resolver serves all three sensors; split across them, power could report DC while current reported AC.

## Notes on the smaller stuff

- **`odometerOnFuelOnly` / `odometerOnBatteryOnly` are lifetime totals in units of 0.1 km.** Both facts come from one invariant — every kilometre was driven on petrol or on battery, so the split must reconstruct the odometer. It does exactly: 63.0 + 33.2 = 96.2 against an odometer of 96.0 on a new car. Read as kilometres they'd claim 630 km of petrol driving in a car that has covered 96 km. Hence a `"deci"` kind in `_coerce` rather than a per-row lambda, and names pairing with Total Mileage rather than the trip-sounding "Distance On …".
- **There is no `distanceToEmptyOnFuel` anywhere in the payload**, so Fuel Range is derived (litres / L-per-100 km). Combined Range stays blank unless both halves are known — showing the electric half alone under that name would read 136 km to a driver with a full tank.
- **`averTraPowerConsumption` was being dropped while its lifetime twin was exposed.** Added as Trip Consumption for every car, not just hybrids — it's an EV field and the asymmetry was the bug.
- **Display precision is derived from the unit in one table via an `_AutoPrecision` mixin.** A constant 15.38 km/kWh graphed as a spike between 15.379999999999999 and 15.380000000000003: the state was already `round(x, 2)`, but the noise comes from HA's 5-minute statistics taking a float mean of N identical samples, after which the chart auto-scales to that 1-ulp range. Rounding harder in `native_value` can't reach it; a display precision can, because statistics honour it. A test fails if any class sets one by hand. Pressure precisions derive from `_PRESSURE_FROM_KPA` rather than being relisted — which is what fixes the four tire-corner sensors, whose unit comes from the setup choice, so psi and bar installs matched nothing and got no precision at all.
- **New sections don't need deleting on a BEV.** Each carries a `visibility` condition on `sensor.<car>_fuel_level`, and a condition on a non-existent entity evaluates false, so they hide themselves. Better than the existing "just delete that tile" note: no per-user edit, can't go stale.
- Two Overview assumptions were battery-electric: it always claimed "Running on battery" (now checks engine speed), and showed electric range alone (now appends fuel range, guarded with `has_value()` so the BEV rendering is unchanged byte for byte).

## Checked against CONTRIBUTING

- **Every numeric sensor has a `state_class`** — audited all 32 specs programmatically, zero united sensors missing one.
- **No `strings.json` / `translations/` churn** — `strings.json` has no `entity` block, so new sensors add no keys, and the five files keep identical key sets.
- **No entry migration needed** — this only *adds* entities; nothing an existing install has is renamed or removed, so no `async_migrate_entry` step and no `VERSION` bump in `config_flow.py`.
- **New entities ship enabled**; Voltage and Current are `DIAGNOSTIC` rather than disabled, and all four raw paths join `_CURATED_PATHS` so the full-exposure pass doesn't hand back a twin of each.
- Pressure stays native kPa with `suggested_unit_of_measurement`, per the "HA converts pressure itself" note.

## Testing

- **`python tests/run.py` at every one of the six commits:** 141 → 143 → 145 → 175 → 204 passed, and the count only ever rises. 39 new tests, each proven to fail against the unfixed source before being trusted.
- **One pre-existing failure throughout**, unrelated to this branch and present at clean `main`: `tests/test_polling_logic.py:114 test_a_parked_car_backs_off_towards_the_cap_but_never_past_it` → `AssertionError: 900.0`.
- **Live verification** on the Starray PHEV: 78 entities; propulsion resolved as `hybrid (declared, tank=True plug=True, powerType='混动')`; fuel range 504 km and combined 640 km against a real diagnostics dump; every display precision in the entity registry matches the table; the three word-valued sensors correctly have none; no `ValueError`; both markdown blocks render through HA's own template engine.

**Heads-up on CI, not caused by this branch:** the Tests job is already red on `main`. It has no `pip install` step, so no manifest requirement is present in the runner and `tests/test_captcha_solver.py` fails 9 tests with `ModuleNotFoundError: No module named 'Crypto'`. I reproduced it identically on a clean `origin/main` worktree. A secondary harness detail hides the cause: `tests/conftest.py:38` registers the module in `sys.modules` *before* `exec_module`, so after the first failure later tests report a misleading `AttributeError: … has no attribute 'fetch_load'` instead of the real import error. Happy to send that as a separate PR if useful.


YossiKon:
@szsolt — this is one of the strongest PRs this repo has received: a second live car on a second backend, six self-contained commits, dozens of tests each proven against the unfixed source, and a description that anticipated nearly every question a reviewer would ask. Thank you. **Merged** (rebase — all six commits preserved), and it ships in v1.12.0.

Before merging I ran a full multi-angle review — propulsion classification, the charge-leg resolver, the metadata refresh, dashboards/YAML, security, and claim-by-claim verification of the PR description. The short version: nearly everything checked out, your two heads-ups were both real bugs and are fixed on `main` with credit to you, and the handful of edge findings the review surfaced are all fixed in follow-up commits rather than sent back to you.

## Verified true (a sample — most claims checked out exactly)

- Both charge-leg counterexamples reproduce: a voltage rule reports the phantom 349 V pack reading forever, and a current-first rule lets the 0.2 A sense current beat a DC fast charge. Those two tests are the most load-bearing in the new suite.
- The 63.0 + 33.2 = 96.2 odometer invariant; the state_class audit (zero misses across every united spec row); translations key parity; the `visibility` trick (missing entity → NaN → false) matches exactly how the frontend evaluates conditions.
- The `_AutoPrecision` MRO mechanics are sound, and "no class sets a precision by hand" is genuinely enforced by the test.

## Your two heads-ups — both real, both fixed (21eb2c6, credited to you)

- **CI red on main**: it installed no dependencies at all. It now installs the manifest requirements **plus Home Assistant**, so the entity tests run in CI instead of skipping.
- **The conftest poisoned-module cache** you diagnosed: exactly right — the entry is now dropped from `sys.modules` on a failed exec, so the real import error surfaces instead of the misleading `AttributeError`.
- Bonus root cause you'll appreciate: the "pre-existing failure" in `test_polling_logic` was a **time-of-day bug**. `_adaptive_interval` has a quiet-hours branch (00:00–06:00 in HA's clock, which is UTC in tests) that returns the cap instead of the base. You opened this PR at 05:24 UTC — that's the only reason you ever saw it fail. The clock is now pinned, and the quiet-hours branch got its own test.

## Findings from the review — all fixed in follow-ups on `main`

None of these were blocking, and none are visible on either test car; flagging them so you can sanity-check my reasoning against the real Starray.

1. **The plug half of the gate wasn't wired** (c72c7e6). `classify()` decided `has_plug`, the tests asserted it — and no platform consumed it, so an HEV would still get all eleven charging entities, the exact "row of dead tiles" the design prevents. `Verdict.charges` now gates them across sensor/binary_sensor/switch/time; UNKNOWN stays permissive so an unreadable payload still degrades to the pre-hybrid set.
2. **Tire display precision** (5d32022). The mixin keyed off the *native* unit, but HA applies a suggested precision to the *display* unit — so the four curated tire sensors handed psi/bar installs kPa's 0 decimals ("2 bar" can't tell a flat tire from a full one), a small regression vs pre-1.12. Now derived from the suggested unit, same single table.
3. **`minutes_or_none` let NaN through** (5d32022) — inherited from the pre-PR inline code, but the helper's contract claimed completeness; `timedelta(minutes=nan)` raises inside the Charge Complete state write. One `math.isfinite` clause.
4. **`_observed_tank` counted a nulled-but-present `fuelStatus` container as a tank** (c72c7e6) — backends send the full schema with blank values; that shape put fuel entities on a BEV whose entry had also lost its powerType.
5. **The refetch guard couldn't heal the entries the old bug damaged** (1bb9feb) — nickname+series truthiness was satisfied by exactly the half-healed entries, locking them out of ever acquiring powerType; and a car with neither nickname nor model re-logged-in on every boot. Key presence now decides, and the heal never downgrades a stored value the server omits on that run.
6. **Full exposure suppressed paths whose curated twin was never built** (c72c7e6) — on a BEV the hybrid fields became invisible everywhere. The skip set is now exactly the entities that exist. Your four charge-leg fields stay suppressed while their computed owners exist, as you designed — with the known trade-off that the parked pack voltage (349 V) is observable nowhere; if you ever want it back as a diagnostic, I'd take that PR.
7. Two small ones (5d32022): the Trip Consumption README row sat in the hybrid-only table though the entity is created for every car; and Fuel Range now reads 0 km on a reported-empty tank with known consumption, so Combined Range doesn't blank exactly when someone is running on the last of both.

Suite on current `main`: **231 passed, 0 failed** — and CI is green again.

If you get a chance to run current `main` on the Starray, I'd love confirmation the 78 entities are unchanged (they should be — every fix is edge-gated). And if you're up for more, the pack-voltage diagnostic or the `timeToTargetDisCharged` consumer would both be welcome follow-ups.

Thanks again. Reviews are easy when the PR does this much of the work.

PR #7, opened 2026-08-04
Only call it charging when the car says it is charging

szsolt:
# Only call it charging when the car says it is charging

Follow-up to #6, on a defect that PR introduced and that live charging exposed.

## The bug

`_charge_leg` picked its leg by the larger `volts × amps`. That is correct while a charge is under way and wrong the rest of the time, because the DC pair is the **pack**, not a charge leg. Driving home unplugged, my car reported:

| `dcChargeUAct` | `dcChargeIAct` | published Charging Power | `Charger Connection` |
|---|---|---|---|
| 340.0 V | 13.2 A | 4.53 kW | Disconnected |
| 345.0 V | 51.3 A | **17.39 kW** | Disconnected |
| 338.2 V | 52.4 A | **17.72 kW** | Disconnected |

Both halves are positive while driving, so the product wins and traction discharge is published as a charge. Over that window the battery went 100 % → 94 % and range 136 → 127 km: the car was emptying while the entity claimed a 17.7 kW charge.

It matters more than a wrong number on a card. `Charging Power` is `device_class: power` with `state_class: measurement`, so those spikes are now **permanently in long-term statistics** for a car whose supply peaks near 1.8 kW, and any "charging above N kW" automation fires on the motorway.

## Why not a smarter rule on the readings

Each candidate picks a different wrong leg, and I tried them in this order:

- **voltage** → picks the pack, which reads 349 V parked and idle;
- **current** → picks the 0.2 A sense noise the AC leg shows while plugged in and not charging;
- **product** → picks traction discharge while driving (the bug above);
- **sign of the DC current** → the pack current only goes negative *once a charge starts* (−4.4 A measured), so it cannot decide whether one has started.

The readings cannot answer the question. The car can, and already does, in `statusOfChargerConnection` — the field that was sitting next to them saying `Disconnected` the whole time. So `_charge_leg` gates on that first, then picks the leg by product exactly as before. The charging codes are **derived** from `_CHARGER_CONNECTION_MAP`:

```python
_CHARGING_CODES = frozenset(
    code for code, label in _CHARGER_CONNECTION_MAP.items() if label == "Charging")
```

so editing a label cannot silently desync the power sensor.

Behaviour: charging → the live leg as before; not charging → `0.0` (idle is a confident zero, not unknown); no charge telemetry at all → unknown, unchanged.

## Pack Power

That leaves the 17 kW with nowhere to go, and it is a real and genuinely useful number — it is what the car's own dashboard shows while driving, which is exactly why I did not want to simply discard it. So it becomes its own entity, `Pack Power`, keeping the car's sign convention: **positive out of the pack, negative in**. It is ungated on the charger, because pack flow is just as true parked as plugged in.

Measured live, charging from a 233 V lead at 7.8 A:

| Entity | Reading |
|---|---|
| Charging Power | 1.83 kW (at the wall) |
| Pack Power | −1.53 kW (into the pack) |
| difference | 0.30 kW — onboard charger + DC-DC + 12 V systems |

One entity cannot serve both roles: `Charging Power` answers "how fast is the wall filling it", `Pack Power` answers "which way and how hard is the pack working". Conflating them is what produced the phantom.

It sits behind the same `charges` gate as its siblings from `c72c7e6`, since it reads the same `dcCharge*` fields — without them it could only ever be unknown.

## Tests

`tests/test_charge_power.py` had a blind spot that is the direct reason this merged: the "charging" fixtures asserted 7.2 kW on a payload whose own `statusOfChargerConnection` said `Disconnected`. **A car never sends that.** A fixture that contradicts itself cannot catch a rule that reads the field it contradicts. Those fixtures now go through a `_charging()` helper that sets a charging status, and the driving case is pinned with the readings above.

Verified against the unfixed source: the driving regression and the four `Pack Power` tests fail on `origin/main` and pass with this change. Full suite **228 passed, 1 failed, 9 skipped** — the failure is `test_an_unreachable_captcha_host_fails_fast_not_five_times`, which fails identically on unmodified `origin/main` in my environment because `numpy` is not installed there; it is not related to this change.

Also deployed and confirmed on the real car: `charging_power` still reads the wall leg at 1.83 kW mid-charge, `pack_power` reads −1.53 kW, and no errors in the log.

Manifest bumped to 1.13.0 — happy to drop that commit if you would rather bump at release time.


YossiKon:
@szsolt — **merged**, and this is exactly how a defect report should look: you found a bug your own PR introduced, led with the incriminating live data, and disproved every tempting alternative rule before reaching for the field that actually answers the question. The four-rules table (voltage → parked pack, current → sense noise, product → traction discharge, DC sign → only flips after the charge starts) is the whole review in one place. Thank you.

What I verified before merging:

- **The fix is right and minimal.** Gating on `statusOfChargerConnection` first, then product — with absence still mapping to unknown and idle to a confident 0.0 — keeps every prior behavior except the phantom. Deriving `_CHARGING_CODES` from `_CHARGER_CONNECTION_MAP` instead of restating `{"3", 3}` is the correct dependency direction; a label edit can't desync the power sensor.
- **The fixture critique is fair and the fix is honest.** The old "charging" fixtures really did describe payloads the car never sends (live amps with `Disconnected`), which is precisely why the driving case slipped through both of us. `_status()` disconnected-by-default with an explicit `_charging()` is the right shape going forward.
- **Pack Power checks out end to end**: kW resolves to 2 decimals through the precision table (verified by instantiation — 17.72 kW driving, −1.53 kW on your wall-charge payload), `device_class: power` + `measurement` so statistics record, sign convention documented as the car's own, and `None` (not 0) when the fields are absent. Nice side effect: the parked pack telemetry that full exposure used to hide behind the resolver sensors now has a legitimate home.
- **Security pass**: no new logging, no server-controlled strings anywhere new, fixtures stay on the fake VIN. Clean.
- **Full suite on the merge: 238 passed, 0 failed** (your 7 new tests included), CI green.

Two small things I handled in a follow-up (679bee7) rather than bouncing back to you:

1. **Your environment's `test_an_unreachable_captcha_host_fails_fast_not_five_times` failure was my bug, and your report was the fix's origin** — the test stubs `solve()` but still imports `geetest_solver`, which pulls numpy at module level. It now skips with the other solver tests when the deps are missing. Thanks for reporting it precisely enough to fix blind.
2. Added `GeelyPackPowerSensor` to `test_derived_sensors_inherit_the_rule_rather_than_restating_it`'s list, so a refactor dropping the mixin there can't pass the suite.

One trade-off worth recording for posterity: a trim that reported charge volts/amps but never `statusOfChargerConnection` would now read 0 kW mid-charge. I think that's the right side to err on — the phantom hits every driver on every drive, permanently, while that trim is hypothetical — and if one ever shows up, Pack Power still tells the truth on it. If you disagree, say so and we'll revisit with data.

For anyone who drove on 1.12.0: the phantom spikes are already in long-term statistics and code can't remove them retroactively — Developer tools → Statistics has an outlier-adjustment for exactly this.

This ships as **v1.13.0** immediately, since every day driven on 1.12.0 writes more phantom kW into people's energy history. Verified on your real car and mine could not ask for more.

szsolt:
Thank you — merged within hours, again. That turnaround genuinely changed how this got fixed: because #6 landed so quickly I had it running on the real car the same day, which is the only reason the phantom charge showed up at all. A slower loop and I'd have found it in a statistics graph in a month, with a polluted history behind it.

For the record, the live outcome after this change, charging from a 233–238 V lead at ~7.8 A:

| Entity | Reading |
|---|---|
| Charging Power | 1.83 kW |
| Pack Power | −1.53 kW |
| difference | 0.30 kW — onboard charger + DC-DC + 12 V |

Charging Power tracks the wall leg through the whole session, Pack Power sits negative while charging and swings positive under traction, and the car's own dashboard agrees within rounding. The 17.7 kW is still there — it just lives on the entity that means it now.

One footnote for anyone cleaning up after the old behaviour: `recorder.purge_entities` clears states but **not** long-term statistics. It returns success and the phantom max survives — mine was still showing 37.54 kW afterwards. The websocket command `recorder/clear_statistics` with the `statistic_ids` is what actually removes it.

And thank you for the repo itself. The propulsion verdict deciding once in `__init__` so every platform agrees, the gates being permissive for a BEV so existing installs are untouched, `strings.json` carrying the user-facing text — that structure is why adding PHEV support was a pleasant afternoon instead of a fight, and why the follow-up fix was a small diff in one resolver rather than a hunt through three sensors. Keeping the climate rename display-only, with `object_id` stable, is the kind of detail most projects get wrong and every user feels. Much appreciated.

PR #13, opened 2026-08-05
Stop reading parkComfortState as the parking-comfort switch

szsolt:
## What

`switch.parking_comfort` reads its state from `_state.parkComfortState`. On my car that value is `1` even with parking comfort **off** in the car, so the switch reported `on` from boot and never changed.

This points the state path at an absent field, so `is_on` returns `None` and the switch reports `unknown` instead.

## Why `parkComfortState` isn't the on/off flag

The `_state` payload has two families, and six `*State` fields read `1` simultaneously on a parked car:

```
activateState      = 1     positionUploadState = 1
inhibitionState    = 1     svtState            = 1
liveDetectionState = 1     parkComfortState    = 1
```

No car runs six modes at once. The `*Active` fields look like the real feature flags — `campingModeActive`, `danceModeActive`, `pulseHeatActive` are all `0` while those features are off. Notably **every other switch in `SWITCH_DEFS` keys off an `*Active` field**; `parkComfortState` was the only `*State` one.

## Why `unknown` rather than a different field

There may be no field to point at:

- No `parkComfort*Active` appears anywhere in that car's 35-key `_state`.
- `charge_server_get("4")` — the parking-comfort endpoint marked **AVD-verified** in `const.py`, which nothing currently polls (`__init__.py` only fetches `bizType=6`) — returns a *schedule* and no state:

```json
{"code":"1000","success":true,"message":"Success",
 "data":{"bizType":4,"scheduleList":[],"startTime":"","endTime":"",
         "bw":"","sh":"","soc":"","timerId":"","id":0}}
```

A `bizType=6` read returned the same empty shape, so that's a well-formed "nothing scheduled", not an error.

Across `_state` and the charge-server endpoint there is nothing reporting parking comfort as currently on or off. That would fit a feature **activated from inside the car**, where the cloud offers a schedule to pre-arm it rather than a toggle — a car you switch on from the seat needs no remote activity flag. I can't prove that negative, but it's consistent with `const.py` marking `SERVICE_PARKING_COMFORT = "RSM"` **"UNVERIFIED on this trim"**.

Which raises a design question for you: if parking comfort is schedule-shaped rather than a toggle, a `switch` may be the wrong entity, and something closer to the existing `time.*_scheduled_charging_start/end` pair might fit better. I've kept this PR to the narrow state fix rather than guessing at that.

## Corrections to my first version of this PR

- I called `parkComfortState` an "availability" flag and implied the real `*Active` field was one capture away. The `bizType=4` read above undercuts that — there may be no such field.
- I said "firing RSM start/stop is untouched". The code path is untouched, but `RSM` is marked unverified on this trim, so I should not have implied the write path is known-good. I have not confirmed the switch actually does anything.

## Scope / caveat

Evidence is from **one vehicle** — a P145-J1 (hybrid, EU cloud), HA 2026.7.4, integration 1.16.1. I have not confirmed whether `parkComfortState` behaves the same on an EX5. If it *is* a genuine on/off flag on some model, this trades a wrong `on` for a less-wrong `unknown` there, so you may prefer a model-conditional read.

Happy to attach the redacted diagnostics dump, or to run further read-only checks.

## Testing

- `python tests/run.py` → **489 passed, 0 failed** (42 skipped: solver deps absent locally)
- `switch.py` coverage **100%**, so the 100% gate holds
- Replaced the parking-comfort assertions in `test_generic_switch_state_comes_from_its_declared_path` (it used this switch to prove generic path-walking; `charging` still covers that) and added `test_parking_comfort_reports_unknown_rather_than_trusting_parkcomfortstate`
- Verified live on HA 2026.7.4: switch reads `unknown`, `charging` still correct, 79 entities, entry loaded, no errors
- `manifest.json` bumped 1.16.1 → 1.16.2 per CONTRIBUTING


YossiKon:
Reviewed line by line - this is exactly the kind of PR a maintainer hopes for. The six-simultaneous-`*State`-fields observation is conclusive on its own, the `*Active` cross-check against every other switch definition seals it, and probing the bizType-4 endpoint to rule out a replacement field before choosing `unknown` is more diligence than most fixes get. Security-wise it's a pure read-path change - reviewed and clean.

Your version bump was the only casualty of time: main moved from 1.16.1 to 1.19.0 while this PR was open, so the manifest hunk conflicted and GitHub wouldn't take the branch as-is. Rather than bounce it back to you for a rebase over one line, I've landed `switch.py` and the tests **verbatim** on main with `Co-authored-by: szsolt`, released as **v1.19.1**, and added one test on top: a payload that *does* carry `parkComfortActive` drives the switch to a real on/off - pinning the behavior your field choice was designed to enable.

On your design question: I think you're right that if parking comfort turns out to be schedule-shaped, a schedule entity (like the existing scheduled-charging pair) is the honest model and the switch should go. Keeping this PR narrow was the right call though - that change wants its own verification pass on a car that actually uses the feature.

Closing since the content is merged; the commit credits you. Thank you - and #14 is getting a full reply next.

szsolt:
Thanks — and there's now direct documentary confirmation for the "activated from inside the car" reading, which I'd only been able to argue circumstantially.

`nitaybz/geely-global-ha`, README, under **Limitations and caveats**:

> - **Some features need to be enabled inside the vehicle first.** Parking Comfort requires the user to turn it on from the car's center screen before the app can control it.

That's the negative I said I couldn't prove, stated as a known vehicle behaviour by the project this integration's protocol layer derives from. It fits every observation in the PR: no `parkComfort*Active` in `_state`, and `bizType=4` returning a schedule with no state field. A feature you switch on from the seat has no remote activation flag to report.

Two things worth flagging, though, because they cut against parts of what I wrote:

**1. This strengthens the schedule-entity design question.** The same README documents `bizType=6` (scheduled charging) as a polled endpoint while `bizType=4` is polled by nothing. If parking comfort is in-car-activated with a cloud-side *schedule*, then a `switch` models the wrong thing regardless of which field it reads, and your instinct to give that its own verification pass on a car that uses the feature looks right.

**2. A caveat on your added test.** Pinning `parkComfortActive` → real on/off is the correct shape, but I should be honest that I have never observed that field on a real payload — I chose the name by analogy with `campingModeActive`/`danceModeActive`, not from a capture. If the field turns out to be named something else, that test passes while the switch stays `unknown` forever on every car. It's pinning my inference, not a captured fact. Might be worth a comment saying so, so a future reader doesn't take the name as verified.

**3. Also relevant to error feedback.** That README documents commands being silently ignored by the car while the server still returns success because it merely *acknowledges* them — and that project surfaces server rejections as toast notifications and deliberately does not apply optimistic state, so the UI never shows a state the car didn't reach. Given this integration applies optimistic state after a fire-and-forget `control()`, that may be worth a separate issue. Happy to file one with the details if useful.

Attribution note: `NOTICE.txt` and 15 source headers already credit `nitaybz/geely-global-ha`, so I'm citing the upstream source of the protocol layer here, not an independent second observation.

Issue #31, opened 2026-08-10
New Geely app asks to set up password


DerHerzog7:
Just downloaded and moved in to the new Geely app which now asks you to set up a password. 

Will this affect the authentication and log in of this integration?

scottaki:
Hi! Quick reassurance: the integration currently authenticates against Geely's existing backend using the email + OTP flow - it's independent of the new app and its password prompt. Accounts without a password keep working exactly as before, and setting a password in the new app doesn't affect the integration's login.

Separately, support for the new app's platform is in progress (it's a separate backend, and the new-platform login also supports the email + OTP flow without a password), so existing accounts should keep working the same way when that lands.

Net effect: the password prompt is safe to ignore (or set - either way the integration is unaffected), and nothing about the new app breaks the current integration login.

YossiKon:
@DerHerzog7 short answer: **no, it does not affect this integration at all.** You can set the password or ignore the prompt; either way nothing here changes.

The reason, from the code rather than from reassurance - this integration never had a password to break:

```
loginType:   3      # 3 = email-code
accountType: 2      # 2 = email
account:     <your email>
code:        <the 6-digit code from the email>
```

That is the whole of what setup sends to `/cidpsso/user/v3/login`. The word "password" appears in exactly one place in the codebase: the list of keys that get **redacted** out of logs and diagnostics. There is no password field in the login request, nothing reads one from your config entry, and nothing would start to just because your account now has one.

**What would actually matter**, so you know what to watch for: not a password *existing* on the account, but Geely *withdrawing* the email-code login. As long as `loginType 3` keeps being accepted, this works. If that ever changed, setup would fail at the code step and Home Assistant would raise a re-authentication prompt rather than failing silently.

---

@scottaki is building new-platform support himself - see his reply below. My earlier note here checked whether *this repository* had such work underway and reported that it does not, which was answering a question he had not asked. His statement was about his own work and it stands.

If the new app eventually replaces the backend this integration talks to, that is its own issue - and @scottaki will have got there first.

scottaki:
@YossiKon "In progress" as in I'm working on it now. Stay tuned.

YossiKon:
@scottaki understood, and my apology - I read "in progress" as a claim about this repository and went and checked branches here, which was the wrong thing to check. You were telling me you are building it, and I should have simply asked.

If you need anything from me, I'm here and happy to help.

YossiKon:
[...]

PR #33, opened 2026-08-10
feat: new Geely EM platform support (com.geely.global.em backend)

scottaki:
# feat: new Geely EM platform support (com.geely.global.em backend)

---

## 1. Summary

Geely released a **separate new app** (new package `com.geely.global.em`, new
backend) and is reported asking users to switch before 2026-08-12 (AU date).
This PR adds support for the new platform as a **coexistent backend** with a clean
**in-place migration path** for existing users, without touching the legacy
backend's behavior.

Key facts that shaped everything (each verified live against a real account,
not assumed):

- The new platform has **no OTP-only login** — pre-switch accounts are
  server-forced through a set-password gate.
- Every `*Encrypt` endpoint requires the password **RSA-1024/PKCS#1-encrypted**
  with a key embedded in the app; plaintext is rejected.
- The new app runs **two sessions**: a "self" session (Zeekr IDaaS) and an
  "HF" session (the **old Ecarx platform**). **All vehicle data
  (status/capability/control/vehicle-list) still lives on the old platform**,
  reachable with an HF JWT minted from the new login. The app fuses both.
- Logging into the new platform **kicks the legacy session** (cross-platform
  single-session) — existing HA users will break at migration time no matter
  what; this PR makes the recovery path clean.

---

### 2.1 Platform picker in the config flow
**Why**: the two backends are incompatible (different hosts, signers,
secrets, session models), and the account's state (pre/post switch) is only
discoverable by probing. The user knows which app they use; asking once is the
simplest honest UX. Default = existing backend, so nothing changes for users
who haven't been migrated (yet).

### 2.2 Password login instead of the legacy email+OTP
Forced by the platform:
- `checkUserV2` on the new IDaaS returns `passwordSet`; pre-switch accounts
  (OTP-only) get **no code-login at all** — the app funnels them into a forced
  `addPassword` migration (`captcha/email` op=`addPassword` →
  `captcha/verify` → `auth/completeMigration`, which also collects first/last
  name — there is **no server-side name source**).
- Post-switch login is `loginByEmailEncrypt{email, RSA(password)}`. We
  deliberately do **not** re-implement the migration UI in HA — the official
  app must be used once first (it also re-links the vehicle, which has **no
  API endpoint**). The config flow says so explicitly.

### 2.3 RSA password encryption
The app's `KtxExtendedKt.a1()`: `RSA/ECB/PKCS1Padding`, 1024-bit PKCS#1 public
key embedded in the app (`ub/b.G`), base64 (76-col wrap). Our earlier
plaintext attempts were rejected (`1001`/`4001`) — this is why the "Encrypt"
suffix on the endpoints exists. Replicated byte-exact; unit vectors against
the captured ciphertext shape.

### 2.4 The HF leg — why the integration talks to the "old" platform
The new app exchanges the IDaaS token for per-client TSP codes
(`user/tspCode?tspClientId=…`, **two** clients). Client 1 mints the "self"
session (`ms-user-auth`, identityType 10). Client 2's code is exchanged at
**`POST api.ecloudkr.com/auth/account/session/secure?identity_type=geelynos`**
(body `{"area":"SEA","authCode":…}`) for an **old-platform JWT (2 days)**.
All vehicle endpoints (status/control/capability/vehicle-list) are the old
platform's, and the new platform's own vehicle-list is currently **empty** for
migrated accounts. The app itself runs both sessions and fuses the data — we
mirror that, but only need the HF session for vehicle data.

### 2.5 Signing identity on the HF leg
The app signs HF calls with the **new app's** identity (`X-APP-ID
GEELY-APP-NEW`, app secret `eeec50…`, v1.0 signer canonical identical to the
legacy one, plus device headers, raw JWT as Authorization — no Bearer). We
replicate the app's traffic **byte-exact** (captured with the real app);
signatures were re-verified against the app's own request vectors. Device
headers carry no captured artifacts: the device identifier is
generated per install and the timezone header comes from the HA config.

### 2.6 In-place migration (Reconfigure) instead of delete+re-add
**Why**: the platform re-registered the account's car under a **new VIN**.
Delete+re-add would regenerate entity unique IDs and split history/automations
for any user whose VIN tail changes. Migration instead:
- keeps the entry and its **legacy-format unique_id** (changing it orphans
  the entity registry — worse than the cosmetic mismatch),
- re-stamps `platform=zeekr`, **purges legacy-only keys** (cidpsso token,
  cert/key paths, device IDs),
- re-points `CONF_VIN` + vehicle metadata to the picked vehicle,
- stores fresh tokens.
A hybrid entry can no longer form (the bug that bit in testing: reauth
without the platform stamp → setup ran the legacy path with dead credentials).

### 2.7 Storing the password (consent + encryption)
The HF JWT lives **2 days**; the platform offers no refresh-token path for it
(the app silently re-authenticates with stored credentials — nobody re-types
passwords). To renew automatically we need the password. Design:
- **consent checkbox** in the flow (default ON, clearly labeled); unchecking
  = manual re-auth every ~2 days,
- **at-rest encryption when the user provides a key**: AES-256-GCM via the
  `cryptography` package (a HA core dependency), key read from `secrets.yaml`
  (`geely_password_key`), stored value prefixed `enc:` — the key lives
  **outside** `.storage`, so a copy of `.storage` alone yields nothing,
- **plaintext fallback** when no key exists, with a warning — same posture as
  the integration's existing private-key storage (`.storage/geely_connect/*/key.pem`),
- never logged; wrong key at decrypt → treated as "no password" → normal
  reauth flow.

### 2.8 Auto-renewal (silent, like the app)
On HF expiry or an auth-looking error: password → IDaaS login (RSA) →
client-2 tspCode → session/secure → new HF JWT, then the coordinator persists
the new token + expiry into the entry. Token lifetimes observed:
IDaaS tokenValue **90 days** (`7776000s`), HF JWT **2 days** (`172800s`).

### 2.9 What we deliberately did NOT do
- No migration UI in HA (app-first message instead) — the migration also
  re-links the vehicle in ways that have no API endpoint.
- No phone-number login (the app supports it; we only ship email).
- No CN-region support (the CN build differs materially).
- No changes to the legacy backend's behavior.

---

### Regions (the big one)
- **Only AU (SEA) live-verified.** The IDaaS host (`gateway-pub-hw-em-sg`),
  the HF host (`api.ecloudkr.com`), the `area` value (`"SEA"`), and the app
  secrets are **assumed** identical for EU/KR/ME/LA but **unverified** — the
  legacy platform had per-region hosts and app secrets, and the new app almost
  certainly does too. The client reads per-region constants; only SEA values
  are populated/verified. **A region matrix test pass is required before
  general release.**
- CN explicitly out of scope.

### Login / account states
- `passwordSet` states: pre-switch (2) and post-switch (password set) verified.
  Other states (`passwordSet=1`?) unseen.
- Phone-number login untested (email only).
- The `identity_type` variants (`geelyos` vs `geelynos`) verified for AU;
  `zeekr-israel` exists in code, untested.

### Vehicle layer
- **Door lock/unlock live-verified (2026-08-10)**: fired from HA against the
  HF endpoints and confirmed on the car by the account owner. **Climate and
  windows are still NOT live-tested.** The request body mirrors the legacy
  `RemoteControlRequest` the integration already sends, but the new app's
  control body differs (`command`/`serviceId`/`setting` object) — the exact
  mapping for each remaining command needs a live pass on a real car.
- Vehicle-list on the NEW platform (`ms-app-bff`) returns **empty** for
  migrated accounts. Whether it ever populates (or exists only for
  new-platform-native accounts) is **unknown** — the integration does not
  depend on it.
- The "state" status variant (`/vehicle/status/state/{vin}`) exists and is
  used by the app's home screen; we use the classic
  `/vehicle/status/{vin}` endpoint (same data family the legacy integration
  parses). The `state` variant's schema is **unexamined**.
- Capability endpoint returns `200` but we observed an **empty data payload**;
  the adapter returns `[]` (documented best-effort default, all-features
  fallback). Whether the server populates it for migrated accounts is unknown.
- Only the EX5 tested; model-specific quirks untested.

### Sessions / expiry
- Cross-platform single-session: new-login-kills-legacy **observed**; the
  reverse (legacy re-auth killing the new session) **assumed** — deliberately
  not tested post-migration.
- HF auto-renewal chain: each leg live-verified individually; the **full
  2-day expiry → silent renewal → entry update loop has not yet been observed
  end-to-end** (first natural expiry lands ~2026-08-12/13).
- `refreshToken` from `ms-user-auth` (self leg): obtained, **unused** (purpose
  unknown).
- The app's own 90-day behavior unobserved (renewal bypasses it via password).

### Environment
- Multiuser/shared-vehicle (`ownerShare` bind) untested.

---

## 4. Files changed (outline)

| File | Change |
|---|---|
| `const.py` | platform constant, zeekr keys, `CONF_STORE_PASSWORD` |
| `config_flow.py` | platform picker; zeekr password login step; re-stamping migration path; consent checkbox; password encryption |
| `zeekr_client.py` | **new**: IDaaS client (X-HMAC signer, checkUserV2, RSA password login, tspCode, completeMigration), ms-user-auth leg, HF leg (v1.0 signer + device headers), vehicle methods |
| `zeekr_adapter.py` | **new**: `GeelyApi`-compatible adapter with silent HF renewal |
| `helpers.py` | HF vehicle-metadata spellings; at-rest credential encryption |
| `__init__.py` | coordinator platform selection; HF token persistence |
| `strings.json` / `translations/*` | new step + guidance strings |

`api.py` (legacy) untouched — legacy behavior is byte-identical.

`zeekr_client.py`'s self-test block (oracle vectors + mock gateway) is
maintained locally (in the author's tree) — the shipped module carries no
test harness, capture fixtures, or real VINs.

---

## 5. Testing summary

**Unit / mock** (all green): client self-test — oracle + X-HMAC signature
vectors, mock gateway roundtrip including HF routes with **v1.0 signature
recomputation**, renewal paths; adapter smoke — auth-retry, silent renewal,
control + error mapping. (The client self-test is local dev tooling, not part
of this PR.)

**Live (AU, real account, 2026-08-10)**: full chain password login → tspCode →
HF exchange → vehicle-list (1 car) → status (`code 1000`, live data);
in-place migration of an existing entry (75 entities live, same IDs, history
preserved); fresh-entry create; **door lock/unlock from HA, confirmed on the
car**.

**Not live-tested**: climate/window control commands; renewal expiry loop
end-to-end; any non-AU region.

---

## 6. Risk / legal note

- The work was done by **private-use reverse engineering** of the production
  app: public HTTPS endpoints, embedded public key, and app constants. No
  credentials, PII, or RE methodology are included in this PR (the endpoints
  and flows are described; the capture/emulator tooling is not). Issue #31's
  note deliberately contains no RE details.
- The integration already ships per-region app secrets for the legacy
  platform; the new platform's single app secret follows the same pattern.
- Live testing used the account owner's own vehicle with one careful probe per
  stage — no retries, no spam.

---

## 7. Follow-ups before/after merge

1. Region matrix pass (EU/KR/ME/LA) — hosts, `area`, secrets, signer-v2
   variants.
2. Live control-command pass for the remaining commands (climate, windows)
   and body-shape alignment with the new app's `RemoteControlRequest` (door
   lock/unlock already verified live).
3. Observe one full HF renewal cycle.
4. Decide the "state" status variant + capability payload handling with real
   data.
5. Decide whether to expose phone-number login.


scottaki:
The **Tests** job is failing on the coverage gate: the new-platform code (`zeekr_client.py`, `zeekr_adapter.py`, the zeekr config-flow steps, and the helpers/coordinator branches) have no tests yet, so total coverage drops below the 100% bar.

I'm on it — porting the client's existing self-test into the suite (oracle signature vectors + a mock-gateway roundtrip that already exercises the signers, IDaaS/HF legs, silent renewal and control paths), plus adapter, config-flow and helpers tests, so the gate should go green.

One question while I write them: the workflow comment says "either test it or discuss lowering the bar" — would you prefer the new-platform code to meet the same 100% coverage, or is a carve-out for the experimental platform acceptable? Happy either way; I just want to match the project's preference before I settle on a shape.

For context, what's live-verified on a real AU account so far: full password-login → vehicle chain, in-place migration of an existing entry, and door lock/unlock confirmed on the car. Climate/windows control and non-AU regions currently remain untested (all detailed in the PR body).

YossiKon:
@scottaki first the small thing you actually asked: keep it at 100%, no carve-out needed — and you've already cleared that bar yourself, so nothing to do there. The tests are the right kind, too; the captured-vector signature tests especially.

I've now read the whole PR properly, and I want to lead with this: it's genuinely impressive work. Reconstructing two session models and fusing an IDaaS login with the old platform's JWT — with the signers oracle-tested against live captures — is hard, careful engineering, and your write-up is the most honest and complete I've seen on a PR here. Thank you for it.

There's really one thing to tighten before it can merge, and it's mechanical rather than a design problem: the new secrets need to be known to the redaction machinery. `api.redact()` and the diagnostics pass both match on exact key names, so `zeekr_access_token`, `zeekr_refresh_token`, `zeekr_hf_token` and `zeekr_password` would pass straight through into a diagnostics download — and since a diagnostics bundle is the one file people attach when they ask for help, that matters (the HF JWT alone can drive control, so it's not only a read token).

**I've taken that half off your plate** — those two files are ours and I know them well, so I've already added the four keys to both passes on `main` (with a test), in `0ec17f1`. So after a rebase that side is done; the only piece left for the new modules is the same reflex the legacy `api.py` uses — wrap the response body in `redact()` at each `raise`, because that text can land on the reauth card. Happy to help with that too if it's useful.

A couple of smaller things I noticed while reading, no rush and none of them blockers:

- It *looks* like the country choice doesn't quite reach the IDaaS headers — `ZeekrIdaas` takes `country` but the header builder is called without it, so every call defaults to `AU`. Did I read that right? If so it'd affect non-AU accounts silently.
- I also want to gently walk back something I implied earlier about legacy users being untouched: because the platform picker is on the shared flow, an existing legacy entry may now see a Reconfigure into it as well, and the zeekr→legacy direction looks like it doesn't clear the old platform stamp. Worth a guard, but I may be misreading the flow — tell me if I am.

There are a few other minor notes (a couple of exception sites that log the full server dict, and the HF-token writeback tripping a full reload) — all small, and I'm happy to walk through them with you whenever suits, rather than pile them on here.

One housekeeping note: `main` moved to v1.37.0 today, so you'll want a rebase anyway — a natural moment to fold the redaction wrapping in. Really glad you're building this.

YossiKon:
@scottaki an update on where this landed, and I want to be straight about the sequence so none of it is a surprise.

Yossi decided to bring the new-platform work into `main` now rather than have me hand you a list and wait on a rebase. So it's merged, shipping as **v1.38.0**. On credit, because that's the part that matters: your commit is merged **as-is** (this PR will show as merged), so your authorship stays in the history, and you're named in `NOTICE.txt` and at the head of both `zeekr_client.py` and `zeekr_adapter.py`. The reverse-engineering — the live migration capture, the two-session model, the snc / IDaaS / HF signers, the RSA password path — is your work and it's credited as yours.

What I changed on top is all security/reliability, not protocol:

- Every `raise` in the client folds its response body through `redact()` (non-JSON bodies dropped), since that text reaches the reauth card — and the four new secret keys are in `api._SECRET_KEYS` / `diagnostics._REDACT`, so a diagnostics download can't carry the tokens or the password.
- The account country now rides the IDaaS `Country` / `RegistCountry` headers (it was accepted and then dropped, so every call went out as AU).
- The silent HF-token writeback no longer trips the update listener into a full reload every ~2 days.
- `secrets.yaml` reads moved off the event loop.
- A reconfigure from the new platform back to legacy now clears the platform stamp + zeekr keys, so it can't leave a hybrid entry.
- The platform is labelled experimental, and the setup text + README now state plainly what the stored password is and how it's protected.

Your tests came across intact and I added a few (redaction-on-raise, the country header, the reload guard). The suite is green at 100%.

Two honest notes. The experimental label reflects that it's still live-verified only on your AU account with door lock/unlock — so climate, windows and non-AU are what real testing confirms next, and you're the person best placed to do it. And if you'd already started a rebase to wrap the raises yourself, I'm sorry for the overlap; the intent was to take the security pass off your plate, not the work out of your hands. Thank you for building this — it's a lot, and it's careful.

scottaki:
@YossiKon thank you — for the merge and for the way you handled it. No overlap on the rebase front: I'd parked the branch once the tests went green, so there was nothing half-done on my side — you took the whole security pass off the plate, which is appreciated.

Both catches were real, and I'll own them. The country parameter being accepted and then dropped — every call silently going out as AU — is exactly the bug that only surfaces for someone with a non-AU account; and the zeekr→legacy reconfigure leaving the platform stamp behind would have created the hybrid entry the migration machinery exists to prevent. The redaction pass is the one I should have had from the start, given the diagnostics bundle is what people attach when they ask for help — the four keys in both passes plus the redact-on-raise reflex is the right shape.

The experimental label is fair, and the honest next step is the one you've named: live verification beyond my AU door lock/unlock. I've done climate (pre-condition) now too and will test the remainder of control functions later today. I'm happy to report back (open an issue or thread with results and any quirks worth documenting). If there's anything else you'd like exercised or documented, let me know.

Thanks again for the credit handling, for folding the tests in intact, and for the added coverage on the redaction and reload paths.

Issue #21, opened 2026-08-06
Adaptive polling can back off while actively driving because driving is derived only from speed


alexandrenakagawa:
Summary

_poll_flags() in __init__.py decides whether the car is "driving" purely from basicVehicleStatus.speed > 0:

python
def _poll_flags(d: dict) -> tuple[bool, bool]:
    ...
    try:
        driving = float(basic.get("speed")) > 0
    except (TypeError, ValueError):
        driving = False
    return charging, driving

_adaptive_interval() then only picks the fast interval when driving (or charging) is True; otherwise it applies the idle exponential back-off (base * 2**idle_streak, up to cap).

The problem: speed == 0 is a legitimate, frequent state while still actively driving (stopped at a red light, in traffic, waiting to turn, etc.), not just a "parked" state. Whenever a poll happens to land on one of these moments, driving flips to False for that cycle, and if the following poll(s) also happen to reuse/return a stale snapshot (or simply land on another speed == 0 moment), the idle-streak counter starts climbing and the coordinator's update_interval grows away from fast toward base/cap — right when the car is still on a trip and the entity that most needs fresh data (speed, GPS) is the one starving.

Observed behavior

On a real drive: sensor.speed updated every ~30-90s for the first ~5 cycles, then stopped changing entirely for the remainder of a ~30 minute drive while the car was actually moving around 40 km/h. 

I haven't been able to fully separate "backend/TCU returned a stale snapshot" from "the adaptive interval widened at exactly the wrong moment" from the outside, but both read as plausible from the code, and the second one is fully within the integration's control to fix regardless of what the backend does.

Suggested fix

Derive driving from a signal that doesn't legitimately hit zero while the trip is still in progress. engineStatus (already read for the curated engine_state sensor, same basicVehicleStatus block) seems like a good candidate:

python
def _poll_flags(d: dict) -> tuple[bool, bool]:
    ...
    engine_running = str(basic.get("engineStatus")).lower() in ("engine_running", "running", "1")
    try:
        driving = engine_running or float(basic.get("speed")) > 0
    except (TypeError, ValueError):
        driving = engine_running
    return charging, driving

(Or, if engineStatus itself turns out to be unreliable on some trims/BEVs without a traditional "engine", tripMeter1/odometer delta between polls could be used as an additional OR-condition - any change there also proves the car moved between the last two snapshots even if the instantaneous speed sample was 0.)

[...]

PR #51, opened 2026-08-24
speed: honour speedValidity, publish unknown when the flag is false

### LINKED ISSUE #44: Speed sensor ignores speedValidity
`basicVehicleStatus` carries `speedValidity` next to `speed`, and the integration reads only `speed`. On my parked EX5 the payload is `speed` = 0.0 with `speedValidity` = false, so the two happen to agree.

They won't always. When the flag goes false the speed value is whatever the car last put there, so a stale non-zero reading gets published as real motion. That matters beyond the sensor itself: the card's `driving_lock` and the adaptive-polling driving heuristic from #21 both key off movement, so a bogus speed can hold the controls locked and keep the poll interval short.

Treating `speed` as unknown when `speedValidity` is false would match what you already do for `steerWhlHeatingSts` = 0, where an unhelpful value becomes unknown rather than a confident number.

MsfPablo:
`basicVehicleStatus` carries `speedValidity` next to `speed`, and the integration read only `speed`. On a parked EX5 the payload is `speed = 0.0` with `speedValidity = false`, so the two agree by coincidence — but when the flag goes false the value is whatever the car last put there, so a stale non-zero reading gets published as real motion.

This treats `speed` as unknown when `speedValidity` is explicitly falsy (`false`, `"false"`, `0`, `"0"`), mirroring the convention already used for `steerWhlHeatingSts = 0`, where an unhelpful value becomes unknown rather than a confident number.

A trim that never reports the flag is unaffected — only an explicit falsy value gates the sensor, so the old behaviour is preserved when the field is absent.

Closes #44.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ Thumbs.db
 
 # Browser automation artifacts (page snapshots, console logs)
 .playwright-mcp/
+
+# Serena local project config
+.serena/
--- a/custom_components/geely_connect/sensor.py
+++ b/custom_components/geely_connect/sensor.py
@@ -665,7 +665,17 @@ class GeelySensor(CoordinatorEntity, _AutoPrecision):
         # working lever instead of an untried one.
         if self._value_offset and isinstance(val, (int, float)):
             val = round(val + self._value_offset, 1)
-        if self._key == "charger_connected" and val == "Plugged in"                 and _is_charging(self.coordinator.data or {}):
+        if self._key == "speed":
+            # `speed` is published alongside `speedValidity`; when the flag is
+            # false the value is whatever the car last put there, so a stale
+            # non-zero reading would be shown as real motion. Treat it as
+            # unknown instead, the same convention `steerWhlHeatingSts` = 0
+            # uses for an unhelpful value. A trim that never reports the flag
+            # is unaffected - only an explicit falsy value gates the sensor.
+            valid = _walk(self.coordinator.data or {}, (*_BASIC, "speedValidity"))
+            if valid is not None and str(valid).strip().lower() in ("false", "0"):
+                return None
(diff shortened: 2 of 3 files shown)


YossiKon:
Merged, and thank you - the reasoning was right and the four tests cover the shape properly, including the two cases that matter most: a stale *non-zero* speed, and a trim that never sends the flag keeping its old behaviour.

**I extended it rather than asking you to.** #44 named a second consumer your change does not reach: `_poll_flags` in `__init__.py` reads `basicVehicleStatus["speed"]` directly, so a disowned non-zero reading still counted as motion there. That is the half with teeth - it holds the *fast* poll interval on a parked car, and the card's driving lock follows the same composite, so every button greys out behind a "Driving" banner while the car sits still.

The rule now lives once, in `helpers.speed_is_stale`, and both callers use it - your inline check is that function's body, moved. Two callers acting on one field is how two rules drift apart, and there is a test asserting that neither file names `speedValidity` itself any more.

One boundary I kept deliberately: the guard removes an input from the driving composite, it does not override it. A car whose ignition is on is still driving with a disowned speed field - otherwise this re-opens [#21](https://github.com/YossiKon/geely-connect/issues/21) from the other side, where a car drove for twenty-five minutes reporting `speed` 0.

Two notes for next time, neither worth changing anything over: the PR also carried a whitespace-only edit on the `charger_connected` line and a docstring reword in an unrelated test. Both harmless, but they make a diff harder to read than it needs to be - and yours was otherwise a very easy one to review.

846 tests pass on `main`. #44 closes with the next release.

YossiKon:
**Shipped in v1.44.0**, and [#44](https://github.com/YossiKon/geely-connect/issues/44) is closed. Your first contribution here - thank you.

Issue #32, opened 2026-08-10
Suggestion/Solution to use the official app concurrently


ropeyajn:
## Problem

The current Home Assistant setup flow performs its own CIDP email/OTP login and provisions a separate mTLS client certificate.  This can replace the official Geely app session or lead to repeated reauthentication. I have attached my solution for this This requires the user to use a MITM proxy to get the relevant `userId` 'auth token' from `/cidpsso/user/v3/login'. I have confirmed that this works. 

## Proposed feature

Please add an advanced, opt-in **Import existing session** setup path.

It should accept an already-authorized CIDP session captured from the official app’s login response, rather than requiring Home Assistant to perform its own email/OTP login.

Desired behaviour:

1. Import the existing CIDP session securely.
2. Discover AU vehicles correctly.
3. Store credentials securely and redact them from logs and diagnostics.
4. Keep normal TLS verification and public-key pinning enabled.
5. Provide a reauthentication prompt when the imported session expires.

## Attached proof-of-concept patch

I have attached a ZIP containing the modified `geely_connect` source.

Changes made:

- `geely_connect/config_flow.py`
  - Adds an advanced session-import setup path.
  - Validates the imported token and CIDP `userId`.
  - Handles AU vehicle discovery.

- `geely_connect/api.py`
  - Adds an AU/APAC fallback (your original integration said I had no vehicles listed it it was using cidpcar/vehicleOwner/v2/controlCars):
    `m-lcmsam-kr.geely.com/cidpcar/vehicleOwner/v1/listControlCars`
    when the existing vehicle-list request returns no vehicles.
  - Adds a certificate-provisioning compatibility retry for AU’s
    `8200 / checkValue invalid` response.

- `geely_connect/strings.json`
- `geely_connect/translations/en.json`
  - Add labels, descriptions, and import-specific errors.

## Why this matters
The main benefit of this setup is the ability to still get push notifications from the car. And it gives users an advanced migration path from an already-authorised official-app session, while preserving the integration’s existing security model. 

Well done with this integration. I had already made my own custom integration using this imported token method before this or the geely global integration was released but this integration is very well designed so I have decided to switch to this. I also have graphics for EX5 that i pulled out of the official iOS app that I can share.

ropeyajn:
I have also unlocked ADB on the vehicle (EX5 extended range) and have full access to the headunit and the TCAM module. Let me know if you want me to check something.

YossiKon:
[...]

PR #54, opened 2026-08-27
New platform: read the vehicle from the new gateway when it lives there (fixes #53)

### LINKED ISSUE #53: New platform: vehicle exists only on ms-app-bff — setup reports "no vehicles", and ms-vehicle-status returns 079001
**Environment:** v1.42.0 (diffed against v1.44.0 — `zeekr_client.py` / `config_flow.py` byte-identical), HA 2026.6.4, AU account, Geely EX5 (`appModelCode E22H-GP`, `engineType BEV`), new EM platform (`com.geely.global.em`).

Two findings: a **root cause with a working fix** for the setup half, and a **precisely-isolated blocker** on the runtime half that I think needs your APK knowledge.

---

## 1. Root cause: the car is only on the new platform

`GET /ms-app-bff/api/v4.0/veh/vehicle-list?needSharedCar=true` on `sea-snc-tsp-api-gw.zeekrlife.com` (snc signer, raw ms-user-auth `accessToken`) returns it immediately:

```json
{"code":"000000","msg":"ok","success":true,"data":[{
  "userVehId":76513,"vin":"…0137","isOwner":true,"defaultVehicle":true,
  "appInnerCode":"E22H-GP","appModelCode":"E22H-GP","appYearCode":"E27KDKFH5213",
  "tboxPlatform":"1","bleProt":"1","engineType":"BEV","marketCode":"13",
  "source":"delivery","nickName":"My GEELY","temId":"0000…","vehicleDirection":"2"}]}
```

**The old Ecarx platform genuinely does not have this car.** The HF leg succeeds, but its JWT is a *different user*:

```
hf_token  : {"uid":"501361132","appId":"…-NEW","operator":"GEELY","env":"production"}
ms-user-auth userId: 2410…f933      IDaaS uuid: 1226…0061
IDaaS checkUserV2  : {"status":1,"registerCountry":"AU","registerSite":"old-geely","passwordSet":1}
```

| call (HF identity) | result |
|---|---|
| `/device-platform/api/v4.0/veh/vehicle-list?needSharedCar=true` | `code 1000`, `data.list: []` |
| `/remote-control/vehicle/status/{vin}?userId=…` | **`8060` The vehicle corresponding to the VIN does not exist** |

`8060` (rather than a permission error) means the VIN is unknown to that platform entirely.

### Ruled out (all tested live, not assumed)

- **Timing** — one session polled every 10 s for 6 min, plus 3 full re-logins: empty throughout. The official app shows this car within **5–10 seconds**, so there is no propagation delay.
- **`needSharedCar`** — `true` / `false` / omitted.
- **Area** — HF `session/secure` minted with `SEA, AP, APAC, AU, SEA-AU, OCEANIA, ASIA`: every one returns the same `uid 501361132` with an empty garage. `api.ecloudeu.com` → `8500`; `api.ecloudus.com` → HTTP 500 / `1500`.
- **`identity_type`** — `geelynos` is the only accepted value (`geely` → `1501 登录失败`, all others → `1400 请求参数不合法`).
- **tspCode client id** — every `*CLIENT*` constant in `zeekr_client`; only `HF_CLIENT2` mints a session, always the same car-less uid.
- **Old-platform endpoint spellings** — `v1.0`/`v3.0`/`v5.0` of the vehicle-list path, `userId=`, `vehicleType=0`, `apis.ecloudkr.com`.
- **The legacy CIDP backend** — the OTP arrives and login succeeds, then it also reports no vehicles, so the v1.36.0 KR `listControlCars` fallback does not cover this shape either.

### Fix for the setup half (working here)

A fail-only probe, in the spirit of the #32 KR fallback:

```python
# zeekr_client.ZeekrClient
def list_vehicles_bff(self, path="/ms-app-bff/api/v4.0/veh/vehicle-list",
                      query="needSharedCar=true") -> dict:
    if not self.access_token:
        raise ZeekrAuthError("not logged in (no new-platform session)")
    return self._request("GET", path, query=query, signer="snc")
```

called only when the old-platform garage returns empty. With this, setup completes and the entry is created. Glad to open a PR.

---

## 2. The blocker: `ms-vehicle-status` refuses this app identity

With the car found, the coordinator fails at `vehicle_status: code=8060` because `ZeekrAdapter` reads status from the old platform. So this account needs the new-platform status route — and that route **exists and rejects us**.

I pulled the route table from the app (`strings` over the 16 `classes*.dex`, no decompilation). The relevant new-platform routes:

```
/ms-vehicle-status/api/v1.0/vehicle/status/latest
/ms-vehicle-status/api/v1.0/vehicle/status/{powerMode,vtm,qrvs,carFridge}
/ms-app-bff/api/v1.0/remoteControl/getVehicleState
/ms-app-bff/api/1.0/user-veh/get-permission          <-- note: api/1.0, no "v"
/ms-remote-control/v1.0/remoteControl/control        <-- note: no /api/ segment
/ms-remote-control/v1.0/remoteControl/setSmartTemp
/ms-vehicle-capability/api/v1.0/vehicle/function/model/info
/ms-charge-manage/api/v1.0/charge/{getLatestSoc,getChargingPlan,setChargingPlan,…}
/ms-vehicle-trail/v1.0/journalLog/{trip/listForPage,trackpoint/list,onOff,delete}
/ms-vehicle-defence/v1/fence/{create,update,delete,enable,page}
/ms-tsp-user-vehicle/api/v2/veh/owner/{checkAuth,share-status,authorize/*}
/ms-app-online-manager/api/v1.0/app/hb
```

(Happy to send the full 149-route list.)

### What these calls do with a valid session

| request | result |
|---|---|
| `GET /ms-vehicle-status/api/v1.0/vehicle/status/latest?vin=…&tboxPlatform=1` + `X-VIN-ID` | **`401 079001` `[SDK]此接口未被授权，无法访问!`** |
| same, `POST` | `400 000010 method not allowed` → GET is correct |
| same, `Authorization: Bearer <tok>` | `401 079020 Invalid token` → **raw token is correct** (matches `vehicle-list`) |
| `getVehicleState`, `charge/getLatestSoc`, `checkAuth` | all `401 079001` |
| `GET /ms-app-bff/api/1.0/user-veh/get-permission?vin=…&tboxPlatform=1` | **`000000` success, `"data": []`** |
| `GET /ms-vehicle-capability/api/v1.0/vehicle/function/model/info` | `400 000002 param check failed` — i.e. **authorized**, params unknown |
| `/ms-app-online-manager/api/v1.0/app/hb` (POST) | `400` listing missing fields (device id/type, heartbeat ts/type) — **authorized** |

Notes on what is *not* the problem:

- **Host is right.** `sea-snc-tsp-api-gw.zeekrlife.com` accepts `X-PROJECT-ID: GEELY` and recognises the route. `snc-tsp-api` / `eu-snc-tsp-api-gw` → `00A02 X-PROJECT-ID validation failed`; `kr-gric-tsp-api-gw.geely.com` → `079002 The app information does not exist`; `gateway-pub-hw-em-sg` → APISIX `404 Route Not Found`; `/power-api/` prefix → Spring 404.
- **`X-PROJECT-ID`** variants (`GEELY_EM`, `GEELYEM`, `EM`, `GEELY-EM`, `GEELYSEA`) all still `079001`.
- **`X-VIN-ID`** is the app's VIN header (the only `X-VIN*` string in the dex); adding it doesn't change `079001`.
- **Two packet captures** of the official app — one signing in, one opening the car with live battery on screen — show it contacting **only** `gateway-pub-hw-em-sg.zeekrlife.com`, `sea-snc-tsp-api-gw.zeekrlife.com` and `api.ecloudkr.com`. There is no separate status host.
- The APK I have is the **base split** — `lib/` is empty, so I could not look at `libenv.so`; `assets/key1.enc` / `key2.enc` (129 B each) are present and unread.

### The two questions

1. **Is `079001` an app-credential ACL, and does the vehicle-service group use different signing material than the BFF?** Same `X-APP-ID: GEELYSEACH001M0001` / `AppId: ONEX97FB91F061405` (the only such strings in the dex) is *accepted* for `ms-app-bff/veh/vehicle-list`, `ms-vehicle-capability` and `ms-app-online-manager`, but *rejected* for `ms-vehicle-status`, `ms-remote-control` and `ms-charge-manage`. Since you decrypted the per-`(brand, area, env)` snc secret out of `libenv.so`, is there a second secret / appId for the vehicle-service group?

2. **Or is the empty `get-permission` the real story** — i.e. this account's vehicle carries no service grants after migration, and `079001` is how the gateway expresses that? If so it's a Geely-side account problem, not an integration bug, and worth documenting so the next owner in this shape isn't sent chasing it.

Also, if you happen to know: what parameters does `ms-vehicle-capability/api/v1.0/vehicle/function/model/info` take? `vin`, `modelCode`, `appModelCode`, `seriesCodeVs`, `innerCode`, `vehicleModelCode`, `vehicleType`, `yearCode`, `marketCode`, `tboxPlatform` and combinations all return `param check failed`.

I can run any specific request against this account and paste the redacted response — it reproduces 100 % of the time.

shanehughes-ui:
Fixes the new-platform side of #53, where an account migrated to the new Geely EM app has an empty **old-platform** garage while its car is plainly visible in the app. On such an account setup ends at `no_vehicles`, and an entry forced into existence fails with `8060` (VIN unknown) because the old platform genuinely does not have that VIN.

Everything here is verified against a capture of the official app (AU account, EX5, 2026-08-27) rather than inferred, and is running live on that account now: **61/77 entities populated, entry `loaded`, no errors** — battery 92 %, range 270 km, doors locked, tyre pressures, location, the lot.

## What changed

**1. Garage — fail-only fallback to `ms-app-bff`.** When the old-platform garage comes back empty, ask the new gateway's mirror of the same route. An account that lists cars today cannot reach this branch, in the spirit of the #32 KR fallback.

**2. Status — the new-gateway read.**

```
GET /ms-vehicle-status/api/v1.0/vehicle/status/latest?latest=false&target=new
    x-vin: <opaque per-vehicle token>
    x-api-signature-nonce: <UUID v4>
    Authorization: <bare access token>
```

The nonce on this gateway is a UUID rather than the 32-char random string the other legs use, and both it and `x-vin` sit inside the signed canonical — `_SNC_WHITELIST` already covers them exactly. The **bare** token is what the gateway accepts here; `Bearer <same token>` is refused with `079020 Invalid token` (the app does send `Bearer`, so something else about its request satisfies that path — moot in practice, but worth knowing).

**3. Payload nesting — the one that looks like a dead end.** The new gateway omits the old platform's `vehicleStatus` level:

```
old:  data = {"vehicleStatus": {basicVehicleStatus, additionalVehicleStatus}, "updateTime": …}
new:  data = {basicVehicleStatus, additionalVehicleStatus, "updateTime": …}
```

`sensor.py` and friends read through `_BASIC = ("vehicleStatus", "basicVehicleStatus")`, so with a perfectly good payload in memory **every** entity reads unknown — except the few that read top-level fields, so `sensor.last_updated` cheerfully shows the current time while nothing else works. The wrapper is restored at the adapter boundary without moving anything, so both spellings resolve. This gateway also reports success as `"000000"` rather than `1000`, translated in the same place so `_SUCCESS_CODES` keeps describing the old platform.

**4. `_renew_hf` re-mints both sessions — a bug in its own right.**

```diff
-        self._client.login_hf(token_value)
+        self._client.login_tsp(token_value)
```

It previously renewed only the HF JWT, leaving `access_token` frozen at whatever the config entry held. That token is single-session server-side: the moment the phone app signs in, the stored one is superseded and every new-gateway call answers `079021 The account is currently logged in elsewhere` — **permanently**, because nothing ever replaced it. `login_tsp` re-mints the snc session and calls `login_hf` itself, so one call recovers both. This affects any new-platform user who opens the app, independently of the rest of this PR.

## The one thing that is not general

`x-vin` is not the VIN — it is an opaque 32-byte token. It is byte-stable per vehicle (identical across two separate logins 40 minutes apart), but it is derived inside the app and **I have not reversed that derivation, and did not try to**. So it is exposed as an optional advanced option (`zeekr_enc_vin`, on the Configure dialog, shown only for new-platform entries) for owners who have captured it, rather than something the project computes.

Empty means the entry behaves exactly as before, on the old-platform status path. Items 1, 3 and 4 are unconditional and need no captured value — **and item 4 in particular is worth taking on its own** even if you would rather not carry the rest.

If you would prefer not to have a "capture required" option in the project at all, that is completely reasonable; item 4 and the `ms-app-bff` garage fallback still stand alone, and I will happily split this into separate PRs — or you may prefer to reimplement from the findings, as with #32. No attachment to this being the implementation.

## Tests

`tests/test_zeekr_new_platform.py` — 5 tests against a mock gateway that **recomputes and verifies the snc signature**, covering the status request shape (path, query, bare token, `x-vin`, UUID nonce, signature), the guard when no token is configured, the bff garage read, and the wrapper (including that it leaves old-platform payloads, non-status dicts and `None` untouched).

One unrelated observation while running the suite: `tests/conftest.py` registers its synthetic package as `gc`, which collides with the stdlib `gc` builtin — that is always present in `sys.modules`, so the `if "gc" not in sys.modules` guard never fires, `gc.__path__` is never set, and `pytest tests/` dies at collection with `No module named 'gc.api'`. It reproduces on a clean checkout of `main` here, so it is not from this branch. Renaming the synthetic package (`_gc_geely`, say) would fix it. Happy to send that as a separate PR if useful.

### DIFF
--- a/custom_components/geely_connect/__init__.py
+++ b/custom_components/geely_connect/__init__.py
@@ -50,6 +50,7 @@ from .const import (
     CONF_VEHICLE_SERIES,
     CONF_VIN,
     CONF_ZEEKR_ACCESS_TOKEN,
+    CONF_ZEEKR_ENC_VIN,
     CONF_ZEEKR_HF_EXPIRY,
     CONF_ZEEKR_HF_TOKEN,
     CONF_ZEEKR_PASSWORD,
@@ -407,6 +408,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             # http.client raise on every HF request.
             timezone=hass.config.time_zone or "UTC",
             hf_expiry=int(d.get(CONF_ZEEKR_HF_EXPIRY) or 0),
+            # Options win over entry data so the token can be pasted or
+            # corrected from Configure without re-adding the integration.
+            enc_vin=(entry.options.get(CONF_ZEEKR_ENC_VIN)
+                     or d.get(CONF_ZEEKR_ENC_VIN) or ""),
         )
     else:
         # Entries created before regions were tracked carry no CONF_REGION and
--- a/custom_components/geely_connect/config_flow.py
+++ b/custom_components/geely_connect/config_flow.py
@@ -34,6 +34,7 @@ from .const import (
     CONF_PLATFORM,
     CONF_REGION,
     CONF_ZEEKR_ACCESS_TOKEN,
+    CONF_ZEEKR_ENC_VIN,
     CONF_ZEEKR_HF_EXPIRY,
     CONF_ZEEKR_HF_TOKEN,
     CONF_ZEEKR_PASSWORD,
@@ -524,6 +525,17 @@ class GeelyIntlConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
                 else:
                     raw = await self.hass.async_add_executor_job(
                         client.list_vehicles, client.user_id)
+                    if not raw:
+                        # Fail-only: an account migrated to the new app can
+                        # have an empty old-platform garage while its car is
+                        # listed on the new gateway. An account that lists
+                        # cars today cannot reach this.
+                        raw = await self.hass.async_add_executor_job(
+                            client.list_vehicles_bff)
+                        if raw:
+                            _LOGGER.info(
+                                "garage found on the new-platform "
+                                "ms-app-bff route")
                     self._vehicles = [v for v in raw if _valid_vin(vehicle_vin(v))]
                     if not self._vehicles:
                         errors["base"] = "no_vehicles"
@@ -708,10 +720,25 @@ class GeelyIntlOptionsFlow(config_entries.OptionsFlow):
             # when an entity is first registered, so afterwards the display
             # unit lives in the registry and nothing the integration reports
             # will move it.
+            if CONF_ZEEKR_ENC_VIN in user_input:
+                user_input[CONF_ZEEKR_ENC_VIN] = (
+                    user_input[CONF_ZEEKR_ENC_VIN] or "").strip()
             new_unit = user_input.get(CONF_PRESSURE_UNIT)
             if new_unit and new_unit != current.get(CONF_PRESSURE_UNIT):
                 _apply_pressure_unit(self.hass, entry, new_unit)
             return self.async_create_entry(title="", data=user_input)
+
+        # New-platform vehicles are addressed by an opaque per-vehicle
+        # token rather than the plain VIN, and it cannot be derived from
+        # anything stored here. Offered only where it applies, and only as
+        # an optional field: without it the entry behaves exactly as
+        # before, reading status from the old platform.
+        extra: dict[Any, Any] = {}
+        if current.get(CONF_PLATFORM, DEFAULT_PLATFORM) == PLATFORM_ZEEKR:
+            extra[vol.Optional(
+                CONF_ZEEKR_ENC_VIN,
+                default=current.get(CONF_ZEEKR_ENC_VIN) or "",
+            )] = str
 
         return self.async_show_form(
             step_id="init",
@@ -747,6 +774,7 @@ class GeelyIntlOptionsFlow(config_entries.OptionsFlow):
                     CONF_EXTERIOR_TEMP_OFFSET,
                     default=float(current.get(CONF_EXTERIOR_TEMP_OFFSET) or 0),
                 ): vol.All(vol.Coerce(float), vol.Range(min=-30, max=30)),
+                **extra,
             }),
         )
 
--- a/custom_components/geely_connect/const.py
+++ b/custom_components/geely_connect/const.py
@@ -92,6 +92,11 @@ CONF_ZEEKR_ACCESS_TOKEN = "zeekr_access_token"
 CONF_ZEEKR_REFRESH_TOKEN = "zeekr_refresh_token"
 CONF_ZEEKR_HF_TOKEN     = "zeekr_hf_token"
 CONF_ZEEKR_HF_EXPIRY    = "zeekr_hf_expiry"
+# The new platform addresses a vehicle by an opaque per-vehicle token in
+# the `x-vin` header rather than by the plain VIN. It is stable per car but
+# derived inside the app, so it is supplied by the owner rather than
+# computed; empty means "stay on the old-platform status path".
+CONF_ZEEKR_ENC_VIN      = "zeekr_enc_vin"
 CONF_ZEEKR_PASSWORD     = "zeekr_password"
 CONF_STORE_PASSWORD     = "store_password"
 
--- a/custom_components/geely_connect/strings.json
+++ b/custom_components/geely_connect/strings.json
@@ -81,7 +81,8 @@
         "data": {
           "poll_mode": "Polling mode",
           "pressure_unit": "Tire pressure unit",
-          "full_exposure": "Expose every raw field (adds ~180 diagnostic entities)"
+          "full_exposure": "Expose every raw field (adds ~180 diagnostic entities)",
+          "zeekr_enc_vin": "Vehicle token for the new platform (x-vin) - advanced, leave blank unless you have it"
         }
       }
     }
--- a/custom_components/geely_connect/translations/en.json
+++ b/custom_components/geely_connect/translations/en.json
@@ -81,7 +81,8 @@
         "data": {
           "poll_mode": "Polling mode",
           "pressure_unit": "Tire pressure unit",
-          "full_exposure": "Expose every raw field (adds ~180 diagnostic entities)"
+          "full_exposure": "Expose every raw field (adds ~180 diagnostic entities)",
+          "zeekr_enc_vin": "Vehicle token for the new platform (x-vin) - advanced, leave blank unless you have it"
         }
       }
     }
--- a/custom_components/geely_connect/zeekr_adapter.py
+++ b/custom_components/geely_connect/zeekr_adapter.py
@@ -48,6 +48,32 @@ def _looks_authy(text: str) -> bool:
     return any(h in low for h in _AUTH_HINTS)
 
 
+def _wrap_vehicle_status(data: Any) -> Any:
+    """Give the new gateway's status payload the old platform's nesting.
+
+    Old platform:  data = {"vehicleStatus": {basicVehicleStatus,
+                            additionalVehicleStatus}, "updateTime": ...}
+    New gateway:   data = {basicVehicleStatus, additionalVehicleStatus,
+                            "updateTime": ...}
+
+    Every consumer reads through the "vehicleStatus" level, so without this the
+    entire entity set reads unknown while the payload sits right there - and
+    the few entities that read top-level fields keep working, which makes the
+    symptom look like anything but a missing key. The wrapper is added without
+    moving anything, so both spellings resolve.
+    """
+    if not isinstance(data, dict) or "vehicleStatus" in data:
+        return data
+    if "basicVehicleStatus" not in data and "additionalVehicleStatus" not in data:
+        return data
+    wrapped = dict(data)
+    wrapped["vehicleStatus"] = {
+        k: v for k, v in data.items()
+        if k in ("basicVehicleStatus", "additionalVehicleStatus")
+    }
+    return wrapped
+
+
 class ZeekrAdapter:
     """GeelyApi-shaped wrapper around ZeekrClient for the new platform."""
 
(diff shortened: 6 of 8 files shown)


YossiKon:
This is remarkable work, and the shape of the PR is right: the `x-vin` derivation is deliberately **not** in it - it is an optional owner-supplied field - which is exactly the line I wanted to hold and you drew yourself. The four findings are each well-isolated, the envelope-wrapper diagnosis (everything unknown while `last_updated` cheerfully shows the current time) is the kind of thing that takes hours to see, and the `_renew_hf` fix looks like a genuine standalone bug that would strand any new-platform user who opens the phone app. Thank you.

I reviewed it against `main` and ran the full suite, and there are two concrete blockers before it can land - both fixable, neither about the approach.

**1. The `x-vin` token is not redacted, so it leaks into diagnostics.** `zeekr_enc_vin` is a per-vehicle capability token - with a live session it addresses and controls the car - so it belongs with the other secrets. This project redacts in two independent passes and a new key has to be in **both** lists or it walks through:

- `api.py` `_SECRET_KEYS`, in the normalised form: add `"zeekrencvin"`
- `diagnostics.py` `_REDACT`: add `"zeekr_enc_vin"`

`tests/test_redaction.py` and `tests/test_diagnostics.py` are where a regression would show. Right now a diagnostics download from a migrated-account entry would publish the token in `entry_data`.

**2. The full suite is red on the branch - 5 pre-existing tests fail.** Your new `test_zeekr_new_platform.py` passes, but `python tests/run.py` overall is `852 passed, 5 failed`. They are all in `test_zeekr_adapter.py`, and they are the `_renew_hf` change: the fixture stubs `c.login_hf`, and now that `_renew_hf` calls `login_tsp` the stub is bypassed and the real method runs. The failing set:

```
test_a_200_wrapped_auth_error_triggers_a_silent_renewal
test_silent_renewal_chain_and_token_take
test_authed_retries_once_on_authy_failure
test_authed_retry_non_authy_failure_propagates_not_reauth
test_zeekr_login_with_no_vehicles_shows_an_error
```

The fix is to update the fixture to stub `login_tsp` (and assert it re-mints both sessions), which also documents the behaviour change your comment describes. The renewal fix is very likely correct - it just needs its tests moved with it.

Everything else checked out: no new network host (the mock's `127.0.0.1` aside), the snc nonce is reassigned before both the header and the signature so they agree, and the bare-vs-Bearer split matches your capture.

On how it lands: this touches the authentication path, and the maintainer's practice there - as on #32 - is to reconstruct auth changes line by line rather than merge them wholesale, so please do not read a request-for-changes as the work being doubted. It is careful and capture-verified; the two items above are what stand between it and green.

Two notes for the record. The **command mapping** you captured (`RDU`+`stop` = lock, `serviceParameters` under `setting`, `x-vin`-only addressing) is not in this PR, and leaving it out until more than Lock is captured is the right call - I would not want inferred bodies going to a car's control endpoint either. And `code "000000"` translating to `1000` at the adapter boundary is the right seam for it; keeping the platforms' vocabularies apart there rather than teaching the coordinator a second success code is the cleaner of the two.

YossiKon:
Landed on `main` (`009471b`), reconstructed with both review blockers closed. Thank you - this is a genuinely hard piece of work, and it was carried far enough that reconstructing it was reading, not guessing.

What changed from your branch, both from the review:

1. **`x-vin` is now redacted** in both passes (`api._SECRET_KEYS` + `diagnostics._REDACT`), with a test. A diagnostics download from a migrated-account entry would otherwise have published the token.
2. **The full suite is green.** The five `test_zeekr_adapter.py` failures were the `_renew_hf` change bypassing a fixture that stubbed `login_hf`; the fixture now stubs `login_tsp` and asserts it re-mints *both* sessions, which documents your fix rather than just accommodating it. One config-flow test also needed its fake to answer the new `list_vehicles_bff` probe.

Three tests added on top of yours, all about the safety boundary, since that is what decided whether this could ship without touching existing users:

- an empty token reads the **old path byte-for-byte** (the invariant for every current entry);
- a token present reads the new gateway, translates `000000`, and re-nests the envelope;
- a migrated account with an empty old garage is **found** via the `ms-app-bff` probe.

Credited to you in the commit and it will be in the release notes. Two things deliberately left for later, both your call as much as mine: the **command mapping** (only Lock is captured, and inferred control bodies do not go to a car), and whether the `x-vin` field should ever become anything other than opt-in "capture required" - it stays exactly that for now.

The `_renew_hf` fix is the part I am most glad you found: it was a live bug for every new-platform user independent of migrated accounts, and it would have been very hard to see from the outside.

I have closed the PR as landed-via-reconstruction rather than merging the branch, which is this project's practice on the auth path (same as #32) - no reflection on the code, which was careful. #53 stays open until you have confirmed the reconstructed version reads your car the same way your branch did; a diagnostics download or a "still works" is all I need.

Issue #24, opened 2026-08-07
Exterior temp +10C offset


alexandrenakagawa:
Evidence of a 10°C offset on the Geely EX5 (fully electric) in Brazil. The 23.2°C reading on the external thermometer occurred because I had been holding it in my warm hand beforehand; it was showing exactly 22.3°C before that.

<img width="521" height="430" alt="Image" src="https://github.com/user-attachments/assets/91d67e4c-303f-4393-9a2e-2c1eafe88790" />

<img width="938" height="883" alt="Image" src="https://github.com/user-attachments/assets/76aeca79-084d-4ac2-95e6-c98985fc15ce" />

YossiKon:
Thank you for this - and for photographing the cluster next to the thermometer, which makes it the best evidence anyone has produced on this field.

I read both images carefully. What they show:

| Source | Reading |
|---|---|
| your handheld thermometer | 22.3 °C (23.2 after being held) |
| **the car's own display** | **22 °C** |
| the cloud field this integration reads | **32.0 °C** |

So the car knows the real temperature and reports it correctly on its own screen, while the cloud's `exteriorTemp` is ten degrees above it. That is a sharper finding than anything on [#11](https://github.com/YossiKon/geely-connect/issues/11), where we only ever had the cloud value against reality.

**But your own history graph is the reason I am not going to subtract ten.** There is a single sample around 07:00 that drops to about 10-11 °C before returning to 32. August in Brazil is winter, and 11 °C at dawn is an entirely plausible real temperature - so that sample may be the *correct* one and everything else stale or offset. The reading was also marked "3 hours ago" when you took the screenshot, so the field is not being refreshed with the rest of the payload. One car showing both behaviours is exactly why the -10 °C correction shipped in v1.21.4 was **retracted** a day later: the field does not report one thing with an offset, it reports different things at different times, and a constant makes some cars worse.

It is documented as a known limitation for that reason, on both the Starray (P145) and the EX5 (E245).

[...]

Issue #11, opened 2026-08-05
Outside Temp & Fuel Range Incorrect


ootyty:
Hey, thank you for your effort in getting Geely onto HomeAssistant! What a legend!

I'm connected as a Geely Starray and notice some minor bug.
I suppose this is originally built for EX5 and suspect that's why the fuel range isn't correct on the Starray.

I notice that the fuel range shown in the car is not matching to the actual range.
Also the outside temperature shown on HA is 10°C above what the car sees.

Thank you!!

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/4b3f5406-cccf-4afa-8985-012d384cdf86" />

<img width="4080" height="3072" alt="Image" src="https://github.com/user-attachments/assets/6ac5fd82-614b-447c-9ea4-5cb1c394851d" />

YossiKon:
Thanks for the clear photos - same clock minute on both screens made this easy to reason about.

Two different things are going on:

**Fuel range** (448 km vs the car's 157 km): that one is on the integration, in an interesting way. The EX5's payload has no fuel-range field, so a range gets projected - litres in the tank over the *lifetime* average consumption. On a plug-in hybrid that average is mostly-electric driving, so the projection triples the truth: your 11.2 L over a lifetime 2.5 L/100km gives exactly the 448 km you saw, while the car itself burns about 7 L/100km when it is actually on petrol. As of v1.17.1, if your car's payload reports a fuel range of its own (the Starray may well, since its cluster shows one), that number wins automatically.

**Outside temperature** (+10 C exactly): the integration reads the `exteriorTemp` field as-is, and an EX5 reads correct through the same code - so the Starray is either reporting this field with an offset or has the real value in a different field.

Both halves need one thing to finish: **Settings - Devices & services - Geely Connect - your car - Download diagnostics**, and attach the file here. It is redacted (VIN, tokens, location are stripped) and it shows every raw field your car sends - which tells me whether there is a real fuel-range field to prefer and what raw number sits behind the temperature. With that file both fixes become certain rather than guessed.

ootyty:
[config_entry-geely_connect-01KZ5MBCS3MQ8T6J7K1WQVPKZF.json](https://github.com/user-attachments/files/30736560/config_entry-geely_connect-01KZ5MBCS3MQ8T6J7K1WQVPKZF.json)

Hey that's sounds very promising, here's the diagnostic! Cheers mate!

YossiKon:
[...]

Issue #10, opened 2026-08-04
statusOfChargerConnection never reaches 3 during a real DC fast-charge session — Charging Power, charging switch, and Charger Plug binary sensor all read wrong


andrewmatten:
I captured a full 41-minute DC fast-charge session on an AU-market EX5 (raw `electricVehicleStatus` polled every 90s, cross-checked against the car's own charger display). `statusOfChargerConnection` stayed at `1` ("Plugged in") the entire time — never reached `3` ("Charging") — even while the car was genuinely pulling ~90kW.

Session summary:

<table>
<tr><th>Event</th><th>Time</th><th>statusOfChargerConnection</th><th>dcDcConnectStatus</th><th>chargerState</th></tr>
<tr><td>Before connect</td><td>23:19:39</td><td>0</td><td>0</td><td>0</td></tr>
<tr><td>Just after connect</td><td>23:21:11</td><td>1</td><td>3</td><td>15</td></tr>
<tr><td>Mid-session (peak ~200A)</td><td>23:34:59</td><td>1</td><td>3</td><td>15</td></tr>
<tr><td>Just before unplug</td><td>00:01:03</td><td>1</td><td>3</td><td>15</td></tr>
<tr><td>After unplug</td><td>00:02:35</td><td>0</td><td>0</td><td>0</td></tr>
</table>

`dcDcConnectStatus` and `chargerState` both flip cleanly at connect/disconnect with zero noise — `statusOfChargerConnection` doesn't move off `1` at all during DC charging.

I found this on a different community integration (`geely_global`) first, then checked this repo's code directly rather than assuming it'd carry over — it does. `_is_charging()` in `sensor.py` gates purely on `statusOfChargerConnection == 3`, and that gate feeds three places:
- `GeelyChargePowerSensor` — would report a flat 0 kW for the whole session (the gate short-circuits to `(0.0, 0.0)` before leg selection ever runs)
- `switch.charging` — same field, same code, would show off
- `binary_sensor.py`'s comment says active-charging state is "surfaced by `switch.charging`" — so this is the one gate the whole integration relies on

I also noticed `test_charge_power.py`'s `_charging()` fixture defaults `statusOfChargerConnection` to `"3"` for every charging case, including the DC ones — that assumption doesn't match what this car actually sent during a real DC session.

[...]

PR #55, opened 2026-08-27
New platform: the lock/unlock/climate command mapping

shanehughes-ui:
Rebased: this PR is now **only the lock/unlock command mapping**, stacked on #60 (the remaining remote commands). Everything else that used to be here — find, windows, climate, seat/wheel heat, air-clean, and now rapid warm/cool — moved to #60, which needs no second-owner confirmation and can land on its own.

What's left here is the single-source part you've rightly held: the door **lock / unlock / tailgate / frunk** directions on RDU/RDL. Directions verified against the car's own `centralLockingStatus`, not the service names — RDU+stop **unlocks**, RDL+start **locks**, and mapped the wrong way `lock.lock` silently opens the car. That's the failure mode that makes one capture insufficient; it waits on a second new-platform owner confirming the cycle.

The only commit here now is `9ec38cc`. Diff shows #60's commits too until that lands (same stacking as before).

### DIFF
--- a/custom_components/geely_connect/zeekr_adapter.py
+++ b/custom_components/geely_connect/zeekr_adapter.py
@@ -173,6 +173,25 @@ def _translate_command(service_id: str, command: str,
     """
     p = {q.get("key"): q.get("value") for q in (parameters or [])}
 
+    if service_id == "RDL_2":                       # lock
+        return "RDL", "start", [{"key": "door", "value": "all"}]
+
+    if service_id == "RDU_2":
+        # RDU is a general "open" service selected by target; the command
+        # differs per target. Directions were checked against the car's own
+        # centralLockingStatus, not the names: RDU+stop UNLOCKS, RDL+start
+        # LOCKS. Mapped the other way round lock.lock silently unlocks the
+        # car - single-source, so this waits on a second owner confirming
+        # the cycle before it should be trusted.
+        target = p.get("target")
+        if target == "trunk":                       # tailgate
+            return "RDU", "stop", [{"key": "target", "value": "trunk"}]
+        if target == "hood":                        # frunk (no entity yet)
+            return "RDU", "start", [{"key": "target", "value": "hood"}]
+        if target:
+            return None
+        return "RDU", "stop", [{"key": "door", "value": "all"}]   # unlock
+
     if service_id in ("RCE_2", "RCE"):
         level = p.get("rce.level")
         seat = p.get("rce.heat") or p.get("rce.ventilation")
--- a/tests/test_zeekr_new_platform_commands.py
+++ b/tests/test_zeekr_new_platform_commands.py
@@ -69,3 +69,29 @@ def test_air_clean_maps_to_rcc():
 
 def test_an_uncaptured_service_is_refused():
     assert tc("SOMETHING_NEW", "start", []) is None
+
+
+# ---- lock / unlock -----------------------------------------------------
+# The directions are the held part: verified against centralLockingStatus,
+# not the service names, because RDU+stop UNLOCKS and RDL+start LOCKS. This
+# stays single-source until a second new-platform owner confirms the cycle.
+
+def test_lock_is_rdl_start():
+    assert tc("RDL_2", "start", [{"key": "door", "value": "all"}]) == (
+        "RDL", "start", [{"key": "door", "value": "all"}])
+
+
+def test_unlock_is_rdu_stop():
+    assert tc("RDU_2", "start", [{"key": "door", "value": "all"}]) == (
+        "RDU", "stop", [{"key": "door", "value": "all"}])
+
+
+def test_tailgate_and_frunk_are_rdu_targets():
+    assert tc("RDU_2", "start", [{"key": "target", "value": "trunk"}]) == (
+        "RDU", "stop", [{"key": "target", "value": "trunk"}])
+    assert tc("RDU_2", "start", [{"key": "target", "value": "hood"}]) == (
+        "RDU", "start", [{"key": "target", "value": "hood"}])
+
+
+def test_an_uncaptured_rdu_target_is_refused():
+    assert tc("RDU_2", "start", [{"key": "target", "value": "sunroof"}]) is None


shanehughes-ui:
Two further commits on this branch since it was opened, both from the same live car.

**`d1a6bab` — the status wrapper now MOVES the blocks instead of copying them.** My original left `basicVehicleStatus` / `additionalVehicleStatus` in place alongside the new `vehicleStatus` key, on the reasoning that moving nothing cannot break a reader. That has a cost I missed: every field then has two paths, and the full-exposure sweep skips only curated paths — which all begin `vehicleStatus.` — so each top-level duplicate becomes an extra raw diagnostic entity. Roughly 200 spurious sensors for anyone who turns full exposure on. Moving leaves exactly one path per field; `updateTime` and friends stay put, and an old-platform payload is still returned untouched.

**`2b1f89c` — nine fields promoted from full exposure to curated entities.** The bar is the one your `cabin_humidity` row sets, and it did most of the deciding for me:

- **four individual door locks** — in a capture across a real lock/unlock cycle all four moved with `centralLockingStatus` and with each other (`1/1/1/1` locked, `0/0/0/0` unlocked), so `0` is the unlocked code exactly as for `trunkLockStatus`. In the same capture the tailgate command moved `trunkLockStatus` 1 → 0 → 1 while all four stayed 1, which is also nice independent corroboration of that sensor's polarity.
- **`isCharging` / `isPluggedIn`** — plain JSON booleans, added *beside* the `statusOfChargerConnection`-derived entities rather than replacing them, since that field is the one that never reaches 3 on a DC fast charge (#10).
- **`tripMeter2`**, beside `tripMeter1`.
- **`disChargeIAct` / `disChargeUAct`** (vehicle-to-load), mirroring the charge pair.

What I deliberately did **not** promote, each for a stated reason in the commit: the **fluid levels** (`0` and `3` on this car; which end means "low" is precisely what a fluid warning must not guess), the **tyre pre/temp warnings** (all `0` with four healthy tyres — the *on* code has never been observed, and a safety-shaped sensor that silently never fires is worse than none), **windows** (`"2"` is documented as closed in `helpers.windows_open` but the open codes are not enumerated, and `winPos` has an unverified scale), **`distanceToEmptyOnBattery20Soc/100Soc`** (294 and 58 against a real range of 268 at 91 % — they do not obviously mean what they say), **`dcChargePileUAct`** (399.75 with nothing plugged in) and the **`healthCharging*`** pair (units unknown).

Verified live: all nine populate — door locks `off` (= locked), trip meter 2 at 173.6 km, discharge pair 0.0 A / 0.0 V, both booleans `off` with nothing plugged in. `tests/test_zeekr_adapter.py` still passes in full.

Happy to split any of this out if you would rather take the coordinate fix, the commands and the entities as separate changes — they are independent, and the coordinate fix is the one I would take first since it also mis-decodes microdegrees on the old platform.

shanehughes-ui:
Two corrections to this PR's description, both mine:

**The car is an EX2, not an EX5.** I inferred "EX5" from `appModelCode E22H-GP` early on and repeated it here and in #54; the owner has since told me it is an EX2. Nothing captured changes — the endpoint, envelope, command vocabulary and coordinate bug are all as described — but the model attribution is wrong wherever it appears, including in #54's landed description.

**The four per-door lock entities in `2b1f89c` are weaker than the rest of that commit**, and I would rather say so than let it pass in review. They are unambiguously *lock* state, not door-open state — the payload carries `doorOpenStatus*` separately and with different values, and these four responded to remote lock/unlock while the car sat parked. But in every capture they moved in lockstep with `centralLockingStatus`, so on this car they carry nothing the aggregate `lock.doors` entity does not already say.

The `trunkLockStatus` precedent is the contrast: that one was observed moving **independently** (1 → 0 → 1 on a tailgate command while all four doors stayed 1), which is what earned it a place. Mine has no such evidence. The lock command does send `door=all`, and a parameter named "all" implies per-door targets exist — but that is inference, and inference about door locks is exactly what I got wrong earlier in #53.

The owner would like them kept on the "more stats, another car may differ" argument, which is reasonable for his install. For the project it is your call, and dropping those four while keeping the other five (`isCharging`, `isPluggedIn`, `tripMeter2`, and the two discharge sensors) would not bother me at all.

YossiKon:
Three parts, and they split cleanly by risk - thank you for making them independent.

## Coordinates — landed, with one change

This is on `main` already (`fa3a934`). I took your diagnosis exactly - new gateway sends degrees, the arc-ms divisor wins because `-33.49/3.6e6` is a valid latitude - but not the reversal, because I checked it against the tropics first and it regresses there:

| coordinate | your order `(1.0, 1e6, ARC)` | correct |
|---|---|---|
| São Paulo lat `-84,780,000` (arc-ms, -23.55) | **-84.78** ❌ | -23.55 |
| São Paulo lon `-167,868,000` (arc-ms, -46.63) | **-167.87** ❌ | -46.63 |

A tropical arc-ms value is small enough that `/1e6` lands inside the degree range, so it wins wrongly - and there are Brazilian cars on this tracker (#11, #24). The `1e6` microdegree divisor is what creates that ambiguity, and no platform is known to send microdegrees, so it is gone: a value already within range is degrees, anything larger is arc-ms. That fixes your Null Island with no tropical regression. Credited to you in the commit; the two cases are pinned in tests.

## Read entities — mostly good, and I verified them against a real *old*-platform car

These are ungated, so they appear on every car, which is right for the ones the old platform also sends - and I checked a real EX5 dump: `doorLockStatusDriver` `1`, `tripMeter2` `8.3`, `disChargeIAct`/`disChargeUAct` `0.0` are all there. So the four per-door locks, Trip Meter 2 and the discharge pair are genuine additions backed by old-platform data, not just new-platform ones. Nice.

Two exceptions: `isCharging` and `isPluggedIn` are **absent** from the old-platform payload, so on an existing car they will read unavailable. Not wrong, but worth either gating them to the new platform or accepting the unavailable state deliberately - a maintainer call on the entity contract, and it also shifts the documented entity counts.

## Commands — the one I will not land on my own

The gating is right: `control` only takes the new path when `enc_vin` is set, so no existing user is touched. And the mapping reads carefully. But this is the one place I stop, and it is exactly your own warning: **`RDL`+`start` locks, `RDU`+`stop` unlocks**, and mapped the other way `lock.lock` silently *unlocks the car*. You got it wrong from the first capture and caught it by watching `centralLockingStatus` - which is the right way, and also why one capture on one car is not enough certainty for me to ship a control path where the failure mode is "you pressed lock and the car opened."

So the commands are the maintainer's call to sign off, the same as the auth path on #32/#54 - not a doubt about your work, a doubt about shipping any single-source unlock mapping without a second confirmation. If another new-platform owner captures the same lock/unlock cycle and `centralLockingStatus` moves the same way, that is what turns it from careful into certain.

Where that leaves #55: the coordinate half is in, the read entities are a small entity-contract decision, and the commands wait on a second confirmation or the maintainer's sign-off. I would rather be slow on the one part that can open a car.

shanehughes-ui:
Added `ffbd38e` — the capability translation, which turned out to be buildable from the single catalogue after all, and produced one finding worth more than the code.

**It works, verified live.** 74 catalogue rows → 9 entries → a parsed view of `ac` / `defrost` / `charging` / `find_car` / `lock` / `unlock` / `tailgate` / `window_vent` / `windows` / `seat.heat` / `steering_wheel_heat` / `parking_comfort`, with `seat.heat.positions ['front-left','front-right']` and `unlock.targets ['door','hood','trunk']`. Every one of those is a control I have captured working on that car, so nothing true was lost in translation.

The asymmetric rule is the part I would most like reviewed: **present ⇒ enabled, absent ⇒ "not fitted" only where the catalogue enumerates the feature at all.** Seat-heat positions arrive one code per seat, so a missing seat-vent code is evidence. The climate entry is labelled 空调服务不区分命令 — *"the AC service does not distinguish commands"* — so it cannot speak to defrost, and reading that silence as "no defrost" would remove a control that works. That class is asserted rather than discovered.

**The finding: an accurate catalogue does not currently remove anything.**

```python
# select.py
vent_supported = caps.get("seat.vent.enabled", True)   # absent -> True
```

```python
# cover.py
add_entities([GeelySunshade(...), GeelySunroof(...), GeelyWindows(...)])
```

Consumers read the flags permissively, and `cover.py` does not consult capabilities at all — so `sunroof.enabled` and `sunshade.enabled` are parsed and then never asked. The owner here has an EX2 with neither, and its catalogue says so plainly (no `C_RWS` sunroof entry; the only mention anywhere is `sunroof_automatic_close`, a rain-reminder under `message_box`), yet both cover entities are still created. He has simply disabled them in the entity registry, which is the right fix for one car and none at all for the general case.

I have **deliberately not changed that**. Making absence mean "off" would change behaviour for every old-platform user too, and if any car's catalogue omits a feature it genuinely has, the result is a missing entity rather than a spare one — the worse direction. It is also your design call rather than mine. If you do want it, gating on `raw_count` (present ⇒ trust the catalogue, absent ⇒ permissive) is the shape that seemed safest to me, but I have no car other than this one to test the old-platform half against.

Tests: `tests/test_zeekr_capabilities.py`, 7 cases including the defrost-is-asserted rule and that an unrecognised catalogue stays permissive. Your existing `tests/test_capabilities.py` (19) and `tests/test_zeekr_adapter.py` (13) both still pass unchanged — 51 green in total across the files this branch touches.

Full 74-row catalogue with glosses is on #53 if it is useful as a reference sample.

shanehughes-ui:
Rebased onto `main`, and the coordinate commit is gone from this branch — your `fa3a934` supersedes it. You were right about the tropics and I would not have caught it: I had no arc-ms car to test against and reached for "try the smallest divisor first" because it fixed the one car in front of me. Deleting the microdegree divisor is the better fix, and my two coordinate tests were redundant against yours, so they went with it. `device_tracker.py` is no longer touched here at all.

**The `isCharging` / `isPluggedIn` gate is in** (`a4b1458`). The signal is the x-vin token rather than the entry's platform, because the token is what decides which status endpoint is read — a zeekr entry without one still polls the legacy path and gets the legacy fields — and it is written in the shape of the propulsion gate beside it. The other seven are untouched, with a test pinning that the gate is exactly two entities wide.

**Two things I owe you, both mine.**

**Your PR was red and neither of us saw it.** `test_the_documented_entity_counts_are_the_real_ones` asserts the README's counts against a real build, and my nine entities made them stale — 76/89 documented against 85/98 built. It skips itself unless Home Assistant is importable, it was not importable in my environment, and no CI runs on a fork PR, so it failed silently. I have installed HA 2026.8.3 and run the suite properly: **810 passed, 0 failed**, 81 skipped (all `test_cards_in_a_browser.py`, playwright). README now reads 83 (96 with a tank), and 85 (98) on the new platform — all four figures checked against an actual build rather than arithmetic.

**I had also added 30 uncovered statements** against your `--fail-under=100` gate: the catalogue read, the command POST and `fetch_capabilities` had no test reaching them. `76ffdc1` closes that — the local gateway now verifies a POSTed body's signature too, so the command body is pinned to the capture with `x-vin` inside the signed canonical, both routes are shown refusing to run without a session or a vehicle token, and on the adapter side an unmapped service raises rather than improvising a body.

`zeekr_adapter.py` and `binary_sensor.py` are at 100%. The eight statements still uncovered in `zeekr_client.py` are the same eight `upstream/main` leaves uncovered in my environment — this branch adds none. Coverage totals: baseline `4499 stmts / 17 miss`, this branch `4601 / 17`.

Both of those were exactly the failure mode you have been guarding against on the command half — shipping on evidence I had not actually checked. You were right to hold it, and I would rather you kept holding it until a second owner captures the lock/unlock cycle. Nothing here changes that; the commands stay gated on `enc_vin` and nothing else moves for an existing entry.

YossiKon:
Re-reviewed against `main` after your rebase, and it has grown a lot - the capability parsing is a welcome addition and I read it through: security-clean, same snc signer and `x-vin` addressing as the read path that shipped in v1.46.0, no new host, no secret, read-only. The per-door locks, Trip Meter 2 and the discharge pair I already confirmed against a real *old*-platform dump, and gating the two new-only booleans behind the token is exactly right.

So almost all of this is good to land. The one thing holding it is unchanged and it is the commands: the lock/unlock mapping is still single-source - your car, your capture, which you yourself got wrong the first time and corrected by watching `centralLockingStatus`. That is the right method, and still not enough certainty for me to ship a control path whose failure mode is "you pressed Lock and the car opened."

**Could you split it?** The capability parsing + read entities in one PR lands immediately - it is verified and I would take it today. The commands in a second PR wait on one thing: a second new-platform owner (lanpup is now here and in your exact shape) capturing the same lock/unlock cycle and confirming `centralLockingStatus` moves the same way. That is what turns the mapping from careful into certain.

I would rather not cherry-pick half of a branch you are still editing - a split from your side keeps the history yours and lets the good two-thirds ship without waiting on the one part that can open a car. v1.46.0 is out, so the read path your work built is now released.

shanehughes-ui:
Split done, both rebased onto current `main`:

- **#58 — capability catalogue + read entities.** The half you said you would take today. Five commits, each green on its own, no command code anywhere in it.
- **This PR — commands only.** Two commits, +289/−0, nothing else in it.

I force-pushed this branch rather than opening a third PR so the lock/unlock discussion above stays with the code it is about.

One thing worth flagging in the split, since it touches a commit of yours: the "move rather than copy" change in #58 alters the assertion you landed in `0d73f6e` — the *"nothing is moved, so top-level readers keep working"* line. Moving is the point of that commit (a copy gives every field a second path, and the full-exposure sweep only skips curated `vehicleStatus.` paths, so each duplicate becomes a spare raw entity), and `updateTime` moves with the blocks so `Car Reported At` is unaffected. The commit carries that test change with it rather than leaving a later one to clean up. If you would rather keep the copy and take the rest, that commit is the first of the five and drops cleanly.

On the second confirmation: I have not approached lanpup. You had already answered their report on #32 with the right diagnostic order, and a second voice saying the same thing seemed like noise. If they land on the EM platform and you would like me to ask them for a lock/unlock cycle with `centralLockingStatus` watched across it, say so and I will — it is their car, so it is their call and yours, not mine.

shanehughes-ui:
Rebased onto current `main` now that #60 landed — this PR is down to the **single lock/unlock commit** (+45, adapter mapping + one test), MERGEABLE/clean. Everything else it used to carry (find/windows/climate, PAI wake, movement wake) shipped in #60, so this is now exactly the one thing you're holding for a second owner.

@YossiKon — the only blocker is unchanged: a second new-platform owner confirming a lock/unlock cycle behaves as the mapping assumes (`RDL`+start = **lock**, `RDU`+stop = **unlock**; the risk you rightly flagged is a reversed mapping opening the car).

@lanpup — you're now confirmed on the EM platform (Find works on your car), so you're exactly the second data point this needs, if you're willing. Two ways, lowest-risk first:

1. **Sensor check (no unmerged command sent):** in the official app, **lock** the car, then **unlock** it, and after each watch HA's door-lock sensors / `centralLockingStatus` — confirm they read locked→unlocked (1 → 0) the same way. That validates the field semantics the mapping relies on.
2. **End-to-end (optional, only if you're comfortable and standing at the car):** install this branch, press **Lock** in HA and confirm the car *locks* (not opens), then **Unlock**. Unlock is fully recoverable and you're right there — but it is the definitive test.

Totally your call (and @YossiKon's) — it's your car. If you'd rather not, no worries; this can wait for whoever's up for it. If either check comes back clean, that's the confirmation to unblock it.

YossiKon:
That's the right split of risk - one clarification on what actually lifts the hold: the sensor check (1) validates the *field* direction (`centralLockingStatus` 1 = locked), which is worth having, but it cannot validate the *command* mapping - a reversed RDL/RDU would read perfectly in (1) and still open the car in (2). So (1) is corroboration; (2), the supervised cycle standing at the car, is the confirmation. Your order is safe as described: supervised, each press's failure mode is a no-op or a recoverable unlock, never a car left open.

@lanpup entirely your call, no pressure - the mapping waits happily either way.

lanpup:
@shanehughes-ui 

Testing results:
<html><head></head><body><p><strong>Caught it—unlocked</strong>, at <strong>3:41:30 pm AEST</strong> after refreshing.</p><div><div><div>
Value | Before | After unlock
-- | -- | --
lock.geely_e22h_gp_1166_doors | locked | unlocked
Raw centralLockingStatus | "1" | "0"

</div></div></div><p>Your existing lock entity <strong>does report the change</strong>. You can monitor it directly; Full exposure isn’t needed. The limitation is the automatic polling delay.</p></body></html>

**Confirmed locked** at **3:42:28 pm AEST**, after refreshing.

- `lock.geely_e22h_gp_1166_doors`: `unlocked` → **`locked`**
- `centralLockingStatus`: `"0"` → **`"1"`**

We’ve now verified both directions. Your existing lock entity tracks the status correctly when refreshed.

lanpup:
@shanehughes-ui also pulled your branch , applied and successfully locked and unlocked the car from Home Assistant. 

Thank you for your service! :)

YossiKon:
@lanpup that is the confirmation this has been waiting on since August - thank you for doing it, and for doing it at the car. Two owners, two cars, the cycle behaving as the mapping says: `RDL`+start locks, `RDU`+stop unlocks.

@shanehughes-ui the evidence bar is met. This is the one control in the project whose failure mode is a car that opens when told to lock, so it goes in on its own rather than folded into a release - I'm putting it to @YossiKon for the final word now that the second cycle exists, and it did not go into v1.49.0 (cut minutes ago) deliberately.

YossiKon:
Landed in 218ae66 and released as **v1.51.0**: https://github.com/YossiKon/geely-connect/releases/tag/v1.51.0

@shanehughes-ui @lanpup thank you both - this is the one that was worth being slow about. The mapping is yours, the method that made it trustworthy was yours too (checking directions against `centralLockingStatus` rather than the service names, and correcting yourself once when they disagreed), and the confirmation cycle at the car is what turned it from a careful guess into a shipped control.

One change on the way in: the code comment saying this "waits on a second owner confirming the cycle" was true when written and is not any more, so it now records what actually happened - two AU cars, locking and unlocking as mapped. An unrecognised RDU target still returns `None` rather than guessing a direction on a door.

Verified before landing: head unchanged from the reviewed diff, no dependency, gateway or redaction surface, and the full gate at 961 tests with coverage 100%.

That closes the new-platform command block.

PR #56, opened 2026-08-28
Car Reported At reads unknown on every new-platform car

shanehughes-ui:
`Car Reported At` reads `unknown` on every vehicle on the new platform, and it is the sensor whose whole job is to say the payload has stopped moving.

`updateTime` is the car's own stamp on the snapshot. The old platform nests it inside `vehicleStatus`, which is where `sensor._reported_at` walks for it; the new gateway sends it at the top level and `_wrap_vehicle_status` left it there. The walk finds nothing, so the sensor is blank — while `Last Updated` keeps advancing beside it, because that one is our clock and the polls are succeeding.

**How it surfaced.** The owner on #53 asked why his car still showed as at home. It had left at 07:20 and parked 18 km away at 07:50; at 09:29 the tracker still read `home`, and nothing anywhere in Home Assistant said the data was old. `Car Reported At` would have read 08:06 all morning. That is #24's scenario exactly — a parked Geely stops reporting, the cloud keeps serving the last snapshot, and a poll every ninety seconds republishes it — with the one entity that reveals it switched off.

**The fix** is one key added to the wrapper's list. I have deliberately kept it to that: whether these blocks should be *moved* rather than copied is a separate question with a different blast radius, and it is on #55 — this wants to be cherry-pickable without it.

Verified against a real new-platform car, and the underlying cause of the stale position is separate and not fixed here: `request_position_refresh` is a no-op on that platform, so nothing ever asks the car for a fresh fix. Opening the phone app refreshes it, and a forced poll then reads the correct location — so the decode and the tracker are fine, only the trigger is missing. That is a capture job and I will raise it separately.

Suite: 783 passed, 0 failed, 81 skipped (playwright). Coverage totals are identical to `main`'s in my environment — `4499 stmts / 17 miss` on both — so this adds no uncovered statement.

### DIFF
--- a/custom_components/geely_connect/zeekr_adapter.py
+++ b/custom_components/geely_connect/zeekr_adapter.py
@@ -52,7 +52,7 @@ def _wrap_vehicle_status(data: Any) -> Any:
     """Give the new gateway's status payload the old platform's nesting.
 
     Old platform:  data = {"vehicleStatus": {basicVehicleStatus,
-                            additionalVehicleStatus}, "updateTime": ...}
+                            additionalVehicleStatus, "updateTime": ...}}
     New gateway:   data = {basicVehicleStatus, additionalVehicleStatus,
                             "updateTime": ...}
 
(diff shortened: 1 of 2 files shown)


YossiKon:
Landed on `main` (`0d73f6e`). Verified your diagnosis against a real *old*-platform dump first, because the wrapper's own docstring had it wrong: it showed `updateTime` as a sibling of `vehicleStatus`, but on a real car it lives **inside** `vehicleStatus` next to basic/additional - which is exactly why walking `vehicleStatus.updateTime` works on the old platform and blanked on the new one until the wrapper carried the key in. Fixed the docstring too.

Added an end-to-end test - the stamp reaching `Car Reported At` after wrapping - rather than only asserting the key is present, since the key being in the right place is the whole point.

This is the sharpest of your finds for what it protects: the coordinate bug put the car in the wrong place, but this one hid the fact that the reading was old at all, which is the failure #24 is entirely about. A parked Geely that stops reporting is the common case; an owner seeing `home` three hours stale with no signal is the bad one.

The position-refresh no-op you flagged - `request_position_refresh` doing nothing on the new platform, so nothing ever wakes the car for a fresh fix - is real and separate, and worth its own issue when you have the capture for the PAI-equivalent. Read-only staleness visibility is the honest half to ship first.

YossiKon:
Closing as landed - the fix is on `main` (`0d73f6e`), reconstructed with the docstring corrected and an end-to-end test, the same way #54 landed. Not merging the branch directly is this project's practice on the adapter path, no reflection on the change, which was right.

Thank you - three verified fixes from one migrated account in a week, and this was the one that mattered most, because it hid staleness rather than just misplacing a value.

PR #57, opened 2026-08-28
feat(zeekr): derive x-vin for new-platform vehicles

scottaki:
## Summary

The new Geely EM platform's status endpoint requires an `x-vin` header. The current integration accepts that value as an advanced, owner-supplied option, but the official app derives it locally from the ordinary vehicle VIN.

This PR derives the header value automatically for vehicles discovered through the new-platform garage path. It leaves the existing manual override in place for compatibility and future app-version changes.

## What the app does

The app's vehicle-request interceptor applies this transform:

    ordinary VIN
      -> AES-128-CBC with PKCS#5/PKCS#7 padding
      -> standard Base64 without line wrapping
      -> X-VIN header

The package-level AES material is not an account credential or session token. It is already publicly documented by the independent MIT-licensed `Wysie/zeekr_key_extractor` project and its public issue history. This change does not bypass login, replace the access token, reproduce app attestation, or change TLS validation.

## Behaviour

- A vehicle found through the new `ms-app-bff` garage fallback automatically gets a derived `x-vin` value.
- Vehicles that remain on the existing HF/Ecarx garage continue using the existing old-platform status path; they do not get routed to the new gateway merely because they use the Zeekr platform login.
- An explicitly configured `zeekr_enc_vin` override continues to take precedence, allowing recovery if the app changes its package material.
- Generated and supplied `x-vin` values are excluded from logs and diagnostics.
- No account credentials, real VINs, captured requests, or live vehicle data are added to the repository.

## Testing

- Adds an offline unit vector using an obviously fake VIN only.
- Verifies the expected Base64 output and padding behaviour.
- Verifies that the new-gateway request receives `x-vin` and that the existing old-platform path remains unchanged when the new-platform source marker is absent.
- Runs the existing standalone test suite and static compilation checks.
- No real authentication call or live vehicle/API validation is part of this change.

## Limitations

The AES material is package/version material and may change in a future app release. The explicit override is retained for that reason. A future update that changes the transform should update the versioned constants and the fake vector rather than exposing any user or account data.

## References

- https://github.com/Wysie/zeekr_key_extractor
- https://github.com/Wysie/zeekr_key_extractor/issues/10
- https://github.com/YossiKon/geely-connect/issues/53
- https://github.com/YossiKon/geely-connect/releases/tag/v1.46.0

### DIFF
--- a/custom_components/geely_connect/__init__.py
+++ b/custom_components/geely_connect/__init__.py
@@ -53,6 +53,7 @@ from .const import (
     CONF_ZEEKR_ENC_VIN,
     CONF_ZEEKR_HF_EXPIRY,
     CONF_ZEEKR_HF_TOKEN,
+    CONF_ZEEKR_NEW_PLATFORM,
     CONF_ZEEKR_PASSWORD,
     CONF_ZEEKR_REFRESH_TOKEN,
     DEFAULT_COUNTRY_CODE,
@@ -393,6 +394,11 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         # what Home Assistant now warns about.
         zeekr_password = await hass.async_add_executor_job(
             password_decrypt, hass, d.get(CONF_ZEEKR_PASSWORD) or "")
+        configured_enc_vin = (
+            entry.options.get(CONF_ZEEKR_ENC_VIN)
+            if CONF_ZEEKR_ENC_VIN in entry.options
+            else d.get(CONF_ZEEKR_ENC_VIN)
+        ) or ""
         api = ZeekrAdapter(
             email=d.get(CONF_EMAIL) or "",
             vin=d[CONF_VIN],
@@ -409,9 +415,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             timezone=hass.config.time_zone or "UTC",
             hf_expiry=int(d.get(CONF_ZEEKR_HF_EXPIRY) or 0),
             # Options win over entry data so the token can be pasted or
-            # corrected from Configure without re-adding the integration.
-            enc_vin=(entry.options.get(CONF_ZEEKR_ENC_VIN)
-                     or d.get(CONF_ZEEKR_ENC_VIN) or ""),
+            # corrected from Configure without re-adding the integration. An
+            # explicitly blank option clears a stale data-level override.
+            enc_vin=configured_enc_vin,
+            new_platform_vehicle=bool(d.get(CONF_ZEEKR_NEW_PLATFORM, False)),
         )
     else:
         # Entries created before regions were tracked carry no CONF_REGION and
--- a/custom_components/geely_connect/api.py
+++ b/custom_components/geely_connect/api.py
@@ -130,7 +130,7 @@ _SECRET_KEYS: set = {
     # With a live session it both addresses AND controls the car, so it is a
     # capability secret, not just an identifier - redacted like the tokens
     # above. Normalised form, matched by both passes; see diagnostics._REDACT.
-    "zeekrencvin",
+    "zeekrencvin", "xvin", "encvin",
     # The IDaaS user-center session token as it appears on the wire (the new
     # platform's login responses carry it as `tokenValue`). No path folds a
     # successful login body into a log today, but masking it is a cheap hedge -
--- a/custom_components/geely_connect/config_flow.py
+++ b/custom_components/geely_connect/config_flow.py
@@ -37,6 +37,7 @@ from .const import (
     CONF_ZEEKR_ENC_VIN,
     CONF_ZEEKR_HF_EXPIRY,
     CONF_ZEEKR_HF_TOKEN,
+    CONF_ZEEKR_NEW_PLATFORM,
     CONF_ZEEKR_PASSWORD,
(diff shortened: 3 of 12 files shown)


YossiKon:
Reviewed carefully, and I am deliberately not merging this - not because anything in it is wrong, but because what it decides is the maintainer's to decide, and I would rather set it out cleanly than make the call.

**The code is careful.** No new host, no new dependency (pycryptodome is already in), the owner-supplied override stays, and the redaction lists are *widened* to cover the derived value and its spellings - which most PRs touching a new secret-shaped field forget. Technically it does what it says: `x-vin = base64(AES-128-CBC(pad(VIN), key, iv))` for BFF-discovered vehicles, so a migrated owner gets live data with no capture at all. That is a real and large usability win - it is exactly what @lanpup and every future migrated owner needs.

**The decision it carries is a values one, and it has two honest sides.**

For: the key is a static package-level constant, already public in the MIT-licensed `Wysie/zeekr_key_extractor`, and it is a VIN-obfuscation rather than an account credential - it defeats no pinning or attestation, and the server accepts it from anyone holding a key that is already out there.

Against: it embeds reverse-engineered key material into the repository. This project has declined extracted material before on provenance grounds (the app graphics on #32), and there is a genuine disagreement in the record about what `x-vin` even is - @shanehughes-ui, who captured it live, treated it as an opaque token from the `libsecret`/`libHttpSecretKey` family and *deliberately did not reverse the derivation*, drawing the same line I did. @scottaki's read is that it is plain AES with a public key. Both are experienced here; they cannot both be the whole story, and which framing is right changes whether shipping this is benign or removes a layer the server leaned on.

So this is squarely @YossiKon's call - it is the authentication/addressing path he reconstructs himself, it embeds a key whose provenance and licensing he would want to weigh, and it is the exact trade the project has taken a careful position on before. I have verified it is safe to run and correct to the claim; whether it *should* ship is not mine to decide.

@scottaki nothing here is a criticism of the work - it is complete and it is the most useful single change for migrated owners. It just lands on the one seam where the maintainer signs personally.

shanehughes-ui:
Tested this against a live **AU/SEA** car (Geely EX2, EM platform, `sea-snc-tsp-api-gw`) — sharing a real-vehicle data point since the PR notes it couldn't be validated against a live car.

**The transform is right.** Probing the account's `ms-app-bff/vehicle-list` confirms it returns the plaintext VIN but *not* the x-vin, and the working x-vin is a 32-byte value (17-char VIN → AES-128 → 32 bytes → 44-char base64) — exactly AES-128-CBC + Base64 as described.

**But the bundled key/IV don't reproduce this car's real x-vin.** With `a01a6db985a2f5d4` / `ed446b8b8845013d` the derived value ≠ the captured, working x-vin. I also brute-checked the obvious variants (key/IV swapped, zero IV, ECB, VIN upper/lower) — none match. So on this AU/SEA build the material differs, consistent with the PR's own caveat and Wysie issue #10 (keys are per-app-version and moved to encrypted constants in v1.5.7).

Implications, FWIW:
- A single hard-coded key/IV will silently derive a **wrong** x-vin on other versions/regions, and unlike a manual value there's no signal until every new-gateway call 401s. Keeping the explicit override authoritative (as this PR does) is the right call.
- Consider a **derive → verify the gateway accepts it → only then persist/use** guard, so a wrong key fails closed instead of replacing a working manual value.
- Ideally the AES material becomes region/app-version-aware. I'm going to try to recover the AU/SEA key/IV and will follow up with the verified constants + a real-car test vector if I get them.

Happy to validate against this vehicle.

YossiKon:
Thanks — a live AU/SEA read is the gap the PR flagged, and it sharpens the call.

It confirms the transform (AES-128-CBC → base64), and more usefully that the bundled `a01a…`/`ed44…` don't reproduce your car's working x-vin. That's the strongest argument yet for keeping the owner-supplied override authoritative rather than shipping a single embedded key: a wrong key derives a *wrong* x-vin with no signal until every new-gateway call 401s — a silent regression for exactly the migrated owners this is meant to help. The per-app-version/region reason (Wysie #10) is plausible but neither of us has pinned it down.

Your derive → verify the gateway accepts it → only then persist guard is the right shape if this ever ships. It stays @YossiKon's call whether to embed extracted key material at all — this doesn't change that the code is safe and correct-to-claim, only that a single static key isn't reliable across builds. If you recover the AU/SEA key/IV plus a real test vector, that's very worth having.

shanehughes-ui:
Followed this up in **#69** — same mechanism you found, but with a verify-before-use twist so the wrong-key case (like the AU/SEA build above) fails safe: it derives with each known `(key, iv)`, checks the gateway actually accepts the result, and stores only a verified value (else falls back to the manual option, unchanged). Ships your pair plus the AU/SEA one, validated against a real car. Credited your PR for the mechanism — happy to fold this into #57 instead if you'd prefer.

YossiKon:
Superseded by #69, which landed in c5d1bb2 carrying this PR's mechanism - `VIN -> AES-128-CBC -> base64`, exactly as you found it - plus a per-build key list and a gateway-verify step that the AU/SEA data made necessary (a single pair turned out to be build-specific, yours included). The landing commit credits this PR by name.

Thank you @scottaki - the derivation discovery is yours, and every migrated owner who now sets up without hunting for an `x-vin` capture has this PR to thank for it.

PR #58, opened 2026-08-28
New platform: capability catalogue and the read entities

shanehughes-ui:
Split from #55 as you asked. This is the half you said you would take today: the capability parsing and the read entities, with no command code anywhere in it — `grep` for `control_new_resp` or `_translate_command` returns nothing on this branch. The commands are #55, rebased down to just themselves.

Rebased onto current `main`, so it sits on top of `0d73f6e`.

**Five commits, each green on its own:**

1. **Move the status blocks rather than copying them.** The wrapper currently does `wrapped = dict(data)` and then adds a filtered `vehicleStatus`, so `basicVehicleStatus.*` and `vehicleStatus.basicVehicleStatus.*` both resolve. The full-exposure sweep skips only curated paths, and those all start `vehicleStatus.`, so every duplicate becomes an extra raw diagnostic entity — about two hundred on this car, each a twin of one already on the list. `updateTime` moves with them, so `Car Reported At` keeps working. This changes one assertion in the test you landed with `0d73f6e` — the "nothing is moved, so top-level readers keep working" line — and the commit carries that change with it rather than leaving it for a later one.
2. **Nine fields promoted to curated entities**, the seven you confirmed against a real old-platform dump plus the two booleans.
3. **The capability catalogue**, translated into the shape `capabilities.py` parses.
4. **The two new-only booleans gated behind the x-vin token**, per your review, with a test pinning that the gate is exactly two entities wide and does not sweep up the other seven.
5. **Coverage for the catalogue route**, which had none.

**Verification.** 799 passed, 0 failed, 81 skipped (playwright). Coverage totals identical to `main`'s in my environment — `17` uncovered on both — so this adds no uncovered statement; `zeekr_adapter.py` and `binary_sensor.py` are at 100%. The README's entity counts are asserted against a real build by your own test, and they are updated: 83 on a battery-only car, 96 with a tank, 85 and 98 on the new platform.

On the asymmetric rule in commit 3, since it is the part I would most like a second opinion on: present code ⇒ enabled, absent ⇒ "not fitted" **only where the catalogue enumerates that feature**. Seat-heat positions arrive one code per seat, so a missing seat-vent code is real evidence; the climate entry is labelled 空调服务不区分命令 — "the AC service does not distinguish commands" — so it cannot speak to defrost, and reading that silence as "no defrost" would remove a control that works. An empty or unrecognised catalogue returns `[]`, which is the permissive all-features view, so this can never expose less than today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ polished setup on top.
 ## ✨ Highlights
 
 - 🔒 **Security-first** - verified TLS plus public-key pinning.
-- 📊 **Everything enabled** - all 76 entities are on from the start (89 on a
+- 📊 **Everything enabled** - all 83 entities are on from the start (96 on a
   hybrid), no duplicates, nothing to switch on by hand.
 - 🧮 **Computed extras** - charging power, charge completion time, range at full
   charge and efficiency, none of which the car reports itself.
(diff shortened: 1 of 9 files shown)


YossiKon:
Landed on `main` as v1.47.0 (`9f1f871`), reconstructed with one docstring line fixed - the wrapper's middle paragraph still said "added without moving anything," which the move you added contradicts.

The split was exactly right, and this half is clean: the MOVE-not-copy fix (about two hundred duplicate raw entities gone from full exposure), the read entities verified against a real old-platform dump before shipping ungated, `is_charging`/`is_plugged_in` gated behind the token, and the capability parsing read-only on the same signer. 880 tests green.

Thank you - three landed contributions now, and the commands wait on #55 for one thing only: a second owner confirming the lock/unlock cycle.

PR #59, opened 2026-08-28
New platform: make position refresh work (PAI), and wake only when moved

shanehughes-ui:
This closes the open thread in `request_position_refresh`'s own docstring on `main` — *"until the PAI write is confirmed"* — it's confirmed. It's the split you asked for: **one self-contained "make location work on the new platform" story**, with no lock/unlock mapping in it (`grep _translate_command` returns nothing here). #55 rebases to just the mapping on top of this.

Rebased onto current `main` (v1.47.0).

**Why it's needed.** The new gateway never streams position — the status payload carries the last *located* fix and only moves when something asks the car for one. On the new platform `request_position_refresh` was a no-op, so nothing ever asked. The owner on #53 watched his migrated car read `home` for **ninety-nine minutes** after it had driven 18 km away, with every other value live and nothing saying the location was stale.

**The fix, capture-verified on a real new-platform car (2026-08-29).** The app's map-open locator is `serviceId PAI` with a single `pai=1` parameter, sent through the same control route as any other new-platform command:

```
POST /ms-remote-control/v1.0/remoteControl/control
{"command":"start","serviceId":"PAI","setting":{"serviceParameters":[{"key":"pai","value":"1"}]}}
```

The catalogue advertises it (`C_PAI_1 位置获取服务`, use=Y). The legacy `operation=4` parameter is rejected on this gateway (`037000 parameter is incorrect`), so only `pai=1` is sent — and that error, versus the fake `000000` some endpoints return for anything, is how the shape was pinned. Verified end to end: the ACK returns, and the fix in the status payload advances from stale to current within seconds.

**Two commits, each green on its own:**

1. **The PAI mechanism.** Adds the shared new-platform control transport (`control_new_resp`) and uses it for exactly one thing — the position wake. **PAI *acquires* a position; there is no variant of it that moves or opens the car**, so unlike the lock/unlock mapping (#55) its failure mode is benign. A vehicle with no x-vin token stays a no-op, so nothing changes for any existing entry.

2. **Wake only when moved.** Now that the wake reaches the car and costs 12 V, *when* it fires matters. It fires while driving, once more when the car parks (so the parked spot is always recorded), and again if the odometer shows a trip the integration never saw — but **never on a bare timer**, so a car parked for a fortnight is never woken. A car whose odometer can't be read keeps the old per-cycle cadence, so no car loses coverage. This half is platform-agnostic — it improves the old platform's battery behaviour too.

**Verification.** 816 passed, 0 failed, 81 skipped (playwright). **Coverage 100%** — `__init__.py`, `zeekr_adapter.py` and `zeekr_client.py` all at 100%, TOTAL 0 uncovered, so it clears the `--fail-under=100` gate you restored in `33307ae`. Tests pin the control envelope, both credential guards, the exact PAI request, the adapter routing, and the wake behaviour driven through the real poll loop across a park cycle.

Running on the reporter's own car now; the map self-updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -568,11 +568,19 @@ profile at setup:
 
 | Mode | Charging / driving | Parked (base → cap) | Secondary data | GPS wake |
 |---|---|---|---|---|
-| 🌙 **Super Eco** - barely polls | every 5 min | 30 min → 2 h | every 2nd cycle | every 24th |
-| 🔋 **Eco** - fewest interruptions | every 90 s | 300 s → 30 min | every 6th cycle | every 12th |
-| ⚖️ **Normal** - balanced | every 30 s | 90 s → 15 min | every 4th cycle | every 6th |
-| ⚡ **Live** - freshest | every 15 s | 45 s → 5 min | every 3rd cycle | every 3rd |
+| 🌙 **Super Eco** - barely polls | every 5 min | 30 min → 2 h | every 2nd cycle | when it moves |
+| 🔋 **Eco** - fewest interruptions | every 90 s | 300 s → 30 min | every 6th cycle | when it moves |
+| ⚖️ **Normal** - balanced | every 30 s | 90 s → 15 min | every 4th cycle | when it moves |
+| ⚡ **Live** - freshest | every 15 s | 45 s → 5 min | every 3rd cycle | when it moves |
 | ✋ **Manual** - you sync | never | never | on every sync | on every sync |
+
+**"When it moves"** means the GPS wake fires while the car is driving, once more
+on the poll after it parks - so the parked location is always the last thing
+recorded - and again if the odometer moves without the integration seeing the
+trip. A car that has not moved is never woken to be asked where it is, however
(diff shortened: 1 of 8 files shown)


YossiKon:
Reviewed at scope level, and it is clean: no new host, no secret or redaction change, gated behind the x-vin token so no existing user is touched, and the full suite is green on the branch. Of the three new-platform command PRs this is the lowest-risk - a position wake either fetches a fresh fix or does nothing, there is no destructive failure mode - so it is the first of the block I would expect to land.

It sits behind the same maintainer sign-off as the rest of the new-platform control code, and behind #55 in the queue only because that one carries the mapping that needs a second owner. Not a hold on this PR's own merits.

YossiKon:
Landed as part of #60 in f07d47c - that branch carried this change verbatim (the PAI wake through the new control route, plus the odometer-anchored wake policy), so it ships together with the command block. Full gate green: 924 passed, 0 failed, 0 skipped, coverage 100%. Closing as landed.

PR #60, opened 2026-08-29
New platform: remaining remote commands (find, windows, rapid climate)

shanehughes-ui:
The remaining new-platform remote commands, split out from #55 so they can land **without** waiting on the lock/unlock second-owner confirmation — none of these has that "press lock, the car opens" failure mode. All capture-verified against the official app (2026-08-27..29) on a real new-platform car, and the working ones re-confirmed by the owner pressing them.

Stacks on #59 (uses its `control_new_resp` transport). #55 is now just the lock/unlock mapping on top of this.

| command | service | mapping |
|---|---|---|
| find / horn / lights | `RHL` | value passes through — `horn-light-flash` (Find), `light-flash` (Lights) |
| windows | `RWS` | `target=window` → start=open, stop=close; `ventilate` = crack |
| rapid warm / cool | `PAA` | **separate endpoint** `POST .../remoteControl/setSmartTemp` — object body, not the control route |
| climate / defrost / seat heat / air-clean | `RCE` / `RCC` | as before |

**Two things worth calling out:**

**A real steering-wheel-heat bug, fixed.** The wheel switch fires with no `rce.level`, so the translator's wheel branch (gated behind a level) was skipped and it fell through to a bare pass-through — dropping the `rce.conditioner=5` the captured body always carries. The wheel then heated but **never turned off**, because the car can't tell which conditioner to stop without it. Now handled before the level check, always sending `conditioner=5`. Confirmed on a real car: on and off both work.

**Rapid climate needed a new endpoint.** On the legacy platform, compound warm/cool is a charge-server `bizType=7` write; the new gateway has a dedicated `setSmartTemp` / `PAA` call with an object body instead — so `ZeekrAdapter.rapid_climate` (a `NotImplementedError` on this backend) now builds that. Warm carries the seat-heat block and `sw=true`; cool an empty ventilation list and `sw=false`; the cabin is driven by `ac`+`temp` either way. Both captures reproduced byte-for-byte.

Gated on `enc_vin` throughout, so no existing entry changes. **829 tests, 100% coverage.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -568,11 +568,19 @@ profile at setup:
 
 | Mode | Charging / driving | Parked (base → cap) | Secondary data | GPS wake |
 |---|---|---|---|---|
-| 🌙 **Super Eco** - barely polls | every 5 min | 30 min → 2 h | every 2nd cycle | every 24th |
-| 🔋 **Eco** - fewest interruptions | every 90 s | 300 s → 30 min | every 6th cycle | every 12th |
-| ⚖️ **Normal** - balanced | every 30 s | 90 s → 15 min | every 4th cycle | every 6th |
-| ⚡ **Live** - freshest | every 15 s | 45 s → 5 min | every 3rd cycle | every 3rd |
+| 🌙 **Super Eco** - barely polls | every 5 min | 30 min → 2 h | every 2nd cycle | when it moves |
+| 🔋 **Eco** - fewest interruptions | every 90 s | 300 s → 30 min | every 6th cycle | when it moves |
+| ⚖️ **Normal** - balanced | every 30 s | 90 s → 15 min | every 4th cycle | when it moves |
+| ⚡ **Live** - freshest | every 15 s | 45 s → 5 min | every 3rd cycle | when it moves |
 | ✋ **Manual** - you sync | never | never | on every sync | on every sync |
+
+**"When it moves"** means the GPS wake fires while the car is driving, once more
+on the poll after it parks - so the parked location is always the last thing
+recorded - and again if the odometer moves without the integration seeing the
+trip. A car that has not moved is never woken to be asked where it is, however
+long it sits. (A car whose odometer cannot be read keeps the old per-cycle
+cadence, and **Refresh Data** forces a fix in any mode - which is also the
+answer for a car that is towed rather than driven.)
 
 The mode is **not fixed at setup** - change it any time from **Configure** on
 the device page (see *Changing settings later* below).
@@ -582,8 +590,8 @@ the device page (see *Changing settings later* below).
 Sits between Eco and Manual: the integration keeps updating on its own, but its
 *starting* interval is where Eco tops out, and two quiet polls reach a two-hour
 ceiling - a parked car costs roughly a dozen sessions a day against Eco's
-~fifty. Because the GPS wake-up reaches the car itself, Super Eco fires it at
-most once per 24 cycles - one wake every two days on a parked car - so it is
+~fifty. Because the GPS wake-up reaches the car itself, it fires only when the
+car has actually moved, so a car parked for a fortnight is never woken at all -
 frugal with the car's 12 V battery, not just with your phone-app session.
 
 Whenever you want more than that, **Refresh Data** is the other half of the
--- a/custom_components/geely_connect/__init__.py
+++ b/custom_components/geely_connect/__init__.py
@@ -264,6 +264,23 @@ _ENGINE_RUNNING = frozenset({"engine_running", "running", "on", "1", "true"})
 _STUCK_POLLS = 10
 
 
+def _odometer(d: dict) -> float | None:
+    """The car's cumulative distance, used as proof that it has moved.
+
+    Strictly increasing and present in every payload, which makes it the one
+    field that can answer "is the position we hold still where the car is?"
+    without asking the car - and asking the car costs a wake.
+    """
+    if not isinstance(d, dict):
+        return None
+    add = (d.get("vehicleStatus") or {}).get("additionalVehicleStatus") or {}
+    raw = (add.get("maintenanceStatus") or {}).get("odometer")
+    try:
+        return float(raw)
+    except (TypeError, ValueError):
+        return None
+
+
 def _poll_flags(d: dict) -> tuple[bool, bool]:
     """(charging, driving) from a status dict, tolerant of missing fields."""
     if not isinstance(d, dict):
@@ -465,7 +482,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     _FAILURE_TOLERANCE = 2
     fail_state = {"consecutive": 0}
     # Efficient-polling state: cycle counter, idle streak, last-change signature.
-    poll_state = {"cycle": 0, "idle": 0, "sig": None}
+    poll_state = {"cycle": 0, "idle": 0, "sig": None, "fix_odo": None}
     # Chosen polling profile (Eco / Normal / Live) from setup.
     profile = POLL_PROFILES.get(
         # Options win over data: the polling mode can be changed after setup.
@@ -487,14 +504,28 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         prev = prev if isinstance(prev, dict) else {}
         was_charging, was_driving = _poll_flags(prev)
 
-        # Position wake (PAI) is expensive and actually wakes the car, so we do
-        # NOT fire it every cycle. Only while driving, once every Nth cycle
-        # when parked - or when the user pressed Refresh Data, whose contract
-        # is "everything, now". That last leg matters most in Super Eco, where
-        # the parked cadence is one wake in days and a manual pull is the
-        # advertised way to get a fresh fix on demand.
+        # Position wake (PAI) reaches the car and spends its 12 V, so it fires
+        # only when the fix we hold could be wrong: while driving (the map
+        # follows the car), on the first poll after a drive ends (`was_driving`
+        # reads the PREVIOUS snapshot, so parking still takes one last fix -
+        # where the car is parked is the reading an owner wants), when the
+        # odometer has moved since the last fix (a trip that began and ended
+        # inside one parked back-off window, speed zero on both sides), or on
+        # Refresh Data. It deliberately no longer fires on a bare timer: an
+        # odometer that has not moved is the car saying it is where it was, so a
+        # car parked for a fortnight is never woken. Movement without driving -
+        # towed, shipped - changes no odometer; Refresh Data covers that.
         forced = poll_state.get("force_secondary", False)
-        if was_driving or forced or ((cyc - 1) % _POSITION_EVERY == 0):
+        prev_odo, fix_odo = _odometer(prev), poll_state["fix_odo"]
+        have_proof = prev_odo is not None and fix_odo is not None
+        moved = have_proof and prev_odo > fix_odo
+        # The timer survives only for a car whose odometer we cannot read -
+        # then nothing is proven and the old cadence stands exactly as before,
+        # so no car loses coverage by this change. It also carries the very
+        # first cycle, when there is no previous snapshot to compare.
+        timer_due = (not have_proof) and ((cyc - 1) % _POSITION_EVERY == 0)
+        woke = was_driving or forced or moved or timer_due
+        if woke:
             try:
                 await _call_with_retry(api.request_position_refresh)
             except GeelyAuthError as e:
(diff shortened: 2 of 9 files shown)


YossiKon:
Same review as #59: clean scope - no new host, no secret/redaction change, gated behind the x-vin token, full suite green on the branch. The `find` half is harmless (honk/flash), the `window` half opens the windows remotely, so it is a notch above the position wake but well below lock/unlock in stakes.

Held in the same place as the rest: new-platform control code is the maintainer's sign-off, and each command mapping is a single-owner capture until a second confirms it. Nothing wrong with the code - it is the "who else has driven this" bar the whole command block sits behind.

lanpup:
@shanehughes-ui That means the PR #60 Find Car test failed. Don’t delete or re-add the integration, and don’t change your x-vin yet.
The PR code is correctly attempting to send RHL through the new-platform endpoint using the x-vin header, but Geely’s server is rejecting that identifier for the remote-control service. Since normal vehicle status still works, this is probably an EX2/Australian API difference—not a typo in your ordinary VIN. [PR #60 details](https://github.com/YossiKon/geely-connect/pull/60)
Send the developer this:
Tested PR #59 + #60 on an Australian Geely EX2. Vehicle status and Refresh Data work, but pressing Find Car fails with:
Geely RHL: code=8060 message=The vehicle corresponding to the VIN does not exist
The integration remains connected and continues receiving vehicle status. I have not changed or re-entered the x-vin.

YossiKon:
Thanks for testing on a real car — that's what this needs.

Before chasing an AU/SEA remote-control difference, one thing to rule out: `code=8060 "vehicle does not exist"` is the **old** telematics endpoint's VIN-unknown code — `PUT /remote-control/vehicle/telematics/{VIN}`, keyed on the plaintext VIN the old platform doesn't hold for a migrated car (it's the same 8060 a status read gets there; see the note in `zeekr_adapter.py`). This PR exists to route new-platform commands *around* that, to `/ms-remote-control/.../control` with the `x-vin` header and no plaintext VIN — so a Find Car that actually reached the new route couldn't come back with an old-endpoint 8060.

Which means that 8060 says the press went out the **old** path — #60's new-control branch didn't run for it. Worth confirming #60's code is genuinely loaded (a full HA restart after installing it, not just a reload — a stale module is easy here) and that `zeekr_enc_vin` is set on the entry, the same token your status read already uses. Then retest Find Car. If it still 8060s *after* you've confirmed it's hitting `/ms-remote-control`, that's a real new-endpoint AU/SEA data point and I'd want the full response.

Either way the PR stays held for @YossiKon's sign-off on new-platform control — this is about getting a clean test, not a merge.

shanehughes-ui:
@lanpup thanks for testing on a real EX2 — and @YossiKon's read is right: that `8060` is the **old** telematics endpoint answering, so the press went out the legacy path, not #60's new `/ms-remote-control/.../control` route. It's not an EX2/AU difference — I run this exact mapping on my own AU EX2 and **Find/lights/windows work**. So it's almost certainly a loading/config issue, not the code.

Quick checklist to get a clean test:

1. **Full HA restart** after installing #60 — not just "Reload". A config-entry reload does **not** re-import the Python modules, so a stale module keeps serving the old path (bit us repeatedly). Settings → System → Restart.
2. **Confirm `zeekr_enc_vin` is set** on the entry — it's the same x-vin token your status read already uses. If it's missing, commands fall through to the legacy VIN path and you get exactly this 8060. (Settings → the integration → Configure shows the field.)
3. Then retest Find Car. If it **still** 8060s *after* you've confirmed a restart + a set x-vin, that would be a genuine new-endpoint AU/SEA data point — grab the full response and I'll dig in.

On the x-vin: if setting/maintaining it by hand is the fiddly part, **#69** derives it automatically from your VIN and **verifies the gateway accepts it** before storing (it ships the AU/SEA build's material, validated against a real AU EX2). With #69 you shouldn't need to touch `zeekr_enc_vin` at all — which removes the most likely cause of this 8060.

lanpup:
> @lanpup thanks for testing on a real EX2 — and @YossiKon's read is right: that `8060` is the **old** telematics endpoint answering, so the press went out the legacy path, not #60's new `/ms-remote-control/.../control` route. It's not an EX2/AU difference — I run this exact mapping on my own AU EX2 and **Find/lights/windows work**. So it's almost certainly a loading/config issue, not the code.
> 
> Quick checklist to get a clean test:
> 
> 1. **Full HA restart** after installing [New platform: remaining remote commands (find, windows, rapid climate) #60](https://github.com/YossiKon/geely-connect/pull/60) — not just "Reload". A config-entry reload does **not** re-import the Python modules, so a stale module keeps serving the old path (bit us repeatedly). Settings → System → Restart.
> 2. **Confirm `zeekr_enc_vin` is set** on the entry — it's the same x-vin token your status read already uses. If it's missing, commands fall through to the legacy VIN path and you get exactly this 8060. (Settings → the integration → Configure shows the field.)
> 3. Then retest Find Car. If it **still** 8060s _after_ you've confirmed a restart + a set x-vin, that would be a genuine new-endpoint AU/SEA data point — grab the full response and I'll dig in.
> 
> On the x-vin: if setting/maintaining it by hand is the fiddly part, **#69** derives it automatically from your VIN and **verifies the gateway accepts it** before storing (it ships the AU/SEA build's material, validated against a real AU EX2). With #69 you shouldn't need to touch `zeekr_enc_vin` at all — which removes the most likely cause of this 8060.

Confirmed full restart. Manually viewed xvin in geely integration and also ran:

cd /config
grep -nE 'if service_id == "RHL"|mapped = _translate_command|control_new_resp' \
custom_components/geely_connect/zeekr_adapter.py

The screenshot confirms all three PR #60 code markers are installed:
- RHL translation exists.
- Commands pass through _translate_command.
- They route to control_new_resp.
Combined with:
- zeekr_enc_vin configured: YES
- the full Home Assistant restart you already performed
…the developer’s prerequisites were satisfied. The PR was installed, although the 8060 response still suggests the running command somehow selected the old route.

Ran debug logging and got the following: 2026-09-01 13:08:58.734 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [XXXXXX - removed in case this is an identiifer?] Error during service call to button.press: Geely RHL: code=8060 message=The vehicle corresponding to the VIN does not exist.

Let me know what else youd like me to try. Cheers

YossiKon:
That's a thorough check — thank you. Full restart, `zeekr_enc_vin` confirmed set, and the three code markers present rules out the stale-module case we were worried about, so the branch is installed and the token is there.

What's still open is the one thing that settles it: **which URL that press actually hit.** With the x-vin set, `control()` should send Find Car to the new `/ms-remote-control/.../control` (x-vin header, no plaintext VIN); the legacy path would be `PUT /remote-control/vehicle/telematics/{VIN}` (plaintext VIN in the URL). Both surface the identical `code=8060 message=…` string, and the integration doesn't log the request path today — so the ERROR line you pasted (the final websocket failure) can't tell them apart. The wording "the vehicle corresponding to the **VIN** does not exist" is how the old VIN-addressed endpoint phrases it, but the new gateway could word it the same, so it hints without proving. I'm not going to guess which it was.

@shanehughes-ui — quickest way to turn this into a fact: `control()` in `zeekr_adapter.py` already has `_LOGGER`, right at the branch point. A one-liner in each side (the `if self._client.enc_vin:` new-route branch and the legacy `return` below it) logging which path it took would make a repeat Find Car, with debug on, print exactly which endpoint went out. Since lanpup already runs your branch, he can just pull that and re-test.

If it prints `/ms-remote-control` and still 8060s, that's the genuine new-endpoint AU/SEA data point — and the next thing worth having is a capture of the **official Geely app** pressing Find Car on lanpup's car (request URL, `x-vin` header, body), to compare how his car addresses remote-control against what the branch sends.

Either way #60 stays held for @YossiKon.

shanehughes-ui:
Two things:

**Find Car confirmed working on my own AU EX2** via this branch — the car honks and flashes, no error. So the RHL mapping and the new route are good on a real new-platform car; @lanpup's 8060 is his command going out the **legacy** path, not the new one.

**Added the path logging @YossiKon asked for** (1573527): `control()` now debug-logs which endpoint it takes at each branch — `NEW route POST /ms-remote-control/...` when the x-vin is set, or `LEGACY route PUT /remote-control/vehicle/telematics/<vin>` otherwise (no VIN/x-vin in the log). Coverage stays 100%.

@lanpup — pull this branch again, then:
1. Enable debug: **Settings → Devices & Services → Geely Connect → ⋮ → Enable debug logging** (or add `logger: logs: custom_components.geely_connect: debug` and restart).
2. Press **Find Car**, then grab the `custom_components.geely_connect.zeekr_adapter` line just before the error.

It'll say `NEW route` or `LEGACY route`:
- **LEGACY route** → the x-vin isn't reaching the adapter, so it's routing to the old endpoint (that's the 8060). Most likely your entry isn't actually on the new platform, or the token isn't applied — worth checking `Download diagnostics` for `"platform": "zeekr"` and a redacted `zeekr_enc_vin`. **#69** (auto-derive + gateway-verify the x-vin) is the clean fix if so.
- **NEW route + still 8060** → that's a genuine new-endpoint AU/SEA data point, and the next thing worth having is a capture of the official Geely app pressing Find on your car (request URL, `x-vin` header, body) to compare against what the branch sends.

lanpup:
@shanehughes-ui  ... rather confusing.... I re-pulled the updated PR #60 branch, performed a full HA restart, and retested Find Car. It now works successfully on my Australian EX2—the car honked/flashed with no 8060 error. I couldn’t find the zeekr_adapter debug line in the logs, but successful execution confirms it reached the new command route. Very confused why a new pull solved this if nothing changed other than extra logging. Im meticluous at this stuff but if it was a me thing I aplogise.

Edit - realised the thing i did differntly this time was that I explicity ran commands from the terminal on HA vs via chatting with the LLM (Chatgpt) instructing it to run commands...so i think a me thing,..

Terminal commands as follows:

(
set -e
cd /config

test -d custom_components/geely_connect
test ! -e custom_components/geely_connect_pr60_pre_pathlog_backup

geely_update_tmp="$(mktemp -d)"

wget -O "$geely_update_tmp/pr60.zip" \
"https://github.com/shanehughes-ui/geely-connect/archive/refs/heads/new-platform-remote-commands.zip"

unzip -q "$geely_update_tmp/pr60.zip" -d "$geely_update_tmp"

geely_new="$geely_update_tmp/geely-connect-new-platform-remote-commands/custom_components/geely_connect"

test -d "$geely_new"
grep -q "NEW route" "$geely_new/zeekr_adapter.py"
grep -q "LEGACY route" "$geely_new/zeekr_adapter.py"

mv custom_components/geely_connect \
custom_components/geely_connect_pr60_pre_pathlog_backup

cp -a "$geely_new" custom_components/geely_connect

grep -nE "NEW route|LEGACY route" \
custom_components/geely_connect/zeekr_adapter.py

echo "Updated PR 60 with endpoint path logging successfully"
)

YossiKon:
No apology needed — this is a useful result, and it explains the 8060 cleanly. Your earlier install evidently hadn't fully applied #60's code (consistent with the LLM-run steps vs the explicit terminal re-pull you noted), so `control()` had no new-platform branch and Find Car went out the legacy `telematics/{VIN}` route — which is exactly that 8060. The clean re-pull, with the `grep` checks confirming the new code is in place, put #60's branch in, and now Find Car takes the new `/ms-remote-control` route: honk + flash, no error.

(The reason you couldn't spot the `zeekr_adapter` debug line is almost certainly that the successful path logs at DEBUG and the level wasn't on for that run — no matter, the honk is the proof it reached the new route.)

So RHL / Find Car is now confirmed on a **second** AU EX2 — thank you both. Worth being precise for the record: that's the Find/horn/lights command specifically; the windows and rapid-climate mappings in #60 are still only exercised on @shanehughes-ui's car, and they're the higher-stakes half. #60 stays held for @YossiKon's call on shipping new-platform control, but this is the confirmation the Find command was waiting for.

YossiKon:
Landed on main in f07d47c (this branch squashed, route logging included). Verified before landing: the full branch diff matches what was reviewed - no dependency, gateway or redaction change, commands stay gated behind the x-vin token, an unmapped service still fails loudly - and the CI gate run locally: **924 passed, 0 failed, 0 skipped, coverage 100%** (zeekr_adapter and zeekr_client both at 100%).

With Find confirmed on two Australian cars and windows/lights live on yours, this ships everything except door lock/unlock, which deliberately stays behind #55's second-owner bar. Thank you both - the captures, the live testing and the odometer-anchored wake are careful work.

Closing; #59 is included (this branch was its superset).

lanpup:
@YossiKon @shanehughes-ui  thankyou. HA just needs a flag to see the new version on main.

YossiKon:
It ships with the next tagged release. HACS 2.x is release-oriented, so `main` only reaches you directly if your HACS still offers **main** under ⋮ → Redownload (older versions do; if the dropdown stops at v1.47.0, that's why - not a fault in your setup). This thread will get a note when the release carrying it is out.

YossiKon:
**v1.48.0 is out** with everything from this thread - the command block, the PAI position refresh and the odometer-anchored GPS wake: https://github.com/YossiKon/geely-connect/releases/tag/v1.48.0

@lanpup update through HACS as normal (Update information if it does not show yet), and you can drop the PR-branch copy - the release supersedes it.

PR #61, opened 2026-08-29
Cards: show seat and wheel heat level as a proportional fill

shanehughes-ui:
The seat-heat buttons lit identically whether set to Low or High — the level was only in the small text. This makes the button **fill in proportion to the level**, so the whole row reads at a glance:

| level | fill |
|---|---|
| High | full (identical to the old on-state) |
| Medium | two-thirds |
| Low | one-third |
| Off | empty |

The steering-wheel switch, being on/off, fills fully or not at all.

**How:** a left-to-right gradient driven by a `--lvl` (0..1) custom property, derived from the select option's *position* so it fits whatever levels the entity offers (not hard-coded to three). The button still carries the `on` class as well, so everything that treats it as "active" is unchanged; the new `.lvl` rule is defined after `.on` and wins only the background. No behaviour change — same tap-to-cycle, same commands — purely how the current level is shown. Applies to seat heat, seat cooling, and wheel heat.

Verified in a browser: the gradient resolves to 100% / 66.7% / 33.3% / 0% for High / Medium / Low / Off, and the buttons still report as lit at every active level (so the existing card tests hold). Independent of the backend PRs — based on `main`.

### DIFF
--- a/custom_components/geely_connect/geely-card.js
+++ b/custom_components/geely_connect/geely-card.js
@@ -673,6 +673,22 @@
     /* The level - Low / Medium / High - is the whole point of a seat button, so
      * it gets the strongest treatment on the card. */
     .cbtn.on b { color:${ON_TEXT}; font-weight: 700; }
+    /* Seat / wheel heat show their LEVEL as a proportional fill: full at the
+     * top level (identical to .on), a third and two-thirds of the button
+     * width at the lower ones, nothing when off - so the setting reads at a
+     * glance, not just from the label. --lvl is 0..1. The button carries both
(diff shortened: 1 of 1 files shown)


YossiKon:
Landed on `main` (`f19c4cf`). The proportional fill reads the setting at a glance - nice touch, and the fraction-from-option-index means it fits whatever levels a trim offers.

One thing I added: the browser test the PR was missing. It changed only the card file, and this project pins card behaviour with a Chromium test so a render cannot silently regress - so I wrote one that asserts the `--lvl` fraction and the fill class at High (1), Low (1/3) and Off (none). Cherry-picked over the display-options change (#64) rather than checked out, so both live together.

PR #62, opened 2026-08-29
Cards: user-supplied car photo, and tap-to-expand

shanehughes-ui:
Two card additions, both opt-in and shipping no artwork.

**Bring your own car photo.** The card draws a generic crossover; this adds a `car_image` option so an owner can point it at a photo of their actual car and have it shown instead. The live state is preserved — headlights, tail lights, charge-port light and the open-panel markers are drawn as an overlay *on top* of the image, positioned as percentages (overridable with `car_image_hotspots`) so one coordinate set fits any side-on shot. No images are bundled: the owner supplies a file under `config/www/`, so it survives add-on updates and there's no third-party artwork in the repo. With no `car_image` set, nothing changes — the drawn car is kept exactly as now.

**Tap the car to expand.** On the compact summary card, tapping the car opens the full card in a modal. The dialog is appended to `document.body`, not the card's shadow root — a top-layer modal centres against its containing block, and the dashboard's masonry/section layouts put a `transform` on a card ancestor, which inside the shadow root sends the dialog to a corner; in `document.body` it centres on the viewport. Backdrop-click and Escape dismiss it; it's torn down with the card.

**Why it matters beyond the feature:** it's the clean, general version of a common request, and it means a card customised with a personal image no longer breaks on update — the image lives in the owner's own `www/`, and the machinery is now in the card itself.

**Tests:** browser tests cover a configured image replacing the SVG while keeping the overlay, no-image keeping the drawing, and the summary card's car opening a centred, populated full-card dialog. I verified all three assertions in a real browser as well. README documents both options. Independent of the other open PRs — based on `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,11 +296,26 @@ Everything else is optional:
 | `battery_bold:` | `true` | Set `false` to draw the battery-% headline in a regular weight instead of bold |
 | `charge_time_format:` | `min` | How **Time to full** reads while charging: `min` (e.g. `249 min`) or `hm` for hours + minutes (e.g. `4h 9m`) |
 | `show_parked:` | `false` | Compact card only. `true` adds a **Parked** chip next to Locked while the car is stationary, matching the full card's status |
+| `car_image:` | *none* | Path or URL to your own photo of the car (e.g. `/local/geely/mycar.png` for a file in `config/www/geely/`). It replaces the drawn car while keeping the live overlay - headlights, tail lights, charge port and the open-panel markers. Transparent PNG/WebP, side-on, front to the left works best |
+| `car_image_hotspots:` | *sensible defaults* | Fine-tune where the overlay lights sit on your image, as percentages, if a different crop needs it - e.g. `{headlight: [8, 46], taillight: [94, 37], port: [86, 40]}` |
 
 ```yaml
 type: custom:geely-card
 nav: [apple, here]        # an iPhone with HERE WeGo on it
 nav_travel: walking       # you are walking to the car
+```
+
+**Your own car, and tap to expand.** By default the card draws a generic
+electric crossover. Point `car_image` at a photo of your actual car - colour and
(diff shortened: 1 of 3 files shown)


YossiKon:
Reviewed and it is good - security-clean in the way that matters for a user-supplied image: `car_image` goes through `esc()` so it cannot break out of the `src` attribute, there are no inline handlers (CSP-safe), and a broken path hides the lamps rather than leaving them over a broken-image glyph. The card code cherry-picks onto `main` cleanly.

It needs a rebase, though, and only because your own other PRs landed first: #64 (display options) and #61 (seat fill) are now on `main`, and they touched the same README options table and the end of `test_cards_in_a_browser.py` that this PR does, so those two files conflict. The `geely-card.js` change itself does not. A rebase onto current `main` clears it, and then I will take it. Sorry for the ordering - the smaller card PRs were the ones ready to land without a test to add.

shanehughes-ui:
Rebased onto current `main`. The two conflicts were exactly where you called them — the README options table and the tail of `test_cards_in_a_browser.py` — both from #64 and #61 landing first; I kept both sides (their new rows/tests plus the `car_image` rows and the three car-image/expand tests). `geely-card.js` merged clean, as you expected. It's back to a single commit over three files and now shows mergeable. Full suite green locally (805 passed; the browser tests skip without a browser, verified in the harness as before). Thanks for the review — no worries on the ordering.

YossiKon:
Landed on `main` (`b46078f`). The rebase was exactly right - both sides kept, `geely-card.js` clean.

Ran the half your environment skips: the full browser suite plus the 100% coverage gate. **898 passed, 0 skipped, coverage 100%** - so the three car-image/expand tests actually executed here and the feature renders as claimed. Security held up on the second read too: `esc()` on the URL, no inline handlers, broken-path hides the lamps. Thank you.

PR #64, opened 2026-08-29
Cards: display options (hide name, battery % headline, Parked chip, charge-time format)

shanehughes-ui:
Three small, opt-in card header options — all defaulting to today's layout, independent, and composable. Requested by an owner who wanted a tighter compact card.

| option | card | effect |
|---|---|---|
| `hide_name` | all | removes the vehicle name from the header (keeps the online status dot) |
| `battery_size` | compact | a px size (e.g. `26`) that promotes the battery **percentage to a headline above the range** at that size, instead of the small top-right reading — tunable so it sits between "small" and the range figure |
| `show_parked` | compact | adds a **Parked** chip next to Locked while the car is stationary, matching the full card's "Parked · Locked" |

With all three on, the compact header drops the name, shows a large battery % above the range, and reads Parked · Locked in the chips. Unset, nothing changes.

Kept deliberately **separate from the car-image PR (#62)** — that's one coherent feature (bring-your-own image + tap-to-expand); this is header layout, a different concern, so each stays focused and independently reviewable.

Browser tests cover each option and the untouched default; README documents all three. No behaviour change to any command or state — purely layout. Based on `main`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,8 +291,9 @@ Everything else is optional:
 | `boot:` | by country | `true` labels the tailgate *Boot*, `false` *Trunk* |
 | `rhd:` | by country | `true` draws the driver on the right in the top view, `false` on the left |
 | `name:` | the device name | A different title for this card |
-| `hide_name:` | `false` | `true` removes the vehicle name from the header entirely (the online status dot stays) |
-| `battery_size:` | *unset* | Compact card only. A pixel size (e.g. `26`) shows the battery **percentage as a headline above the range** at that size, instead of small in the top-right - tune it to taste |
+| `hide_name:` | `false` | `true` removes the vehicle name **and the status dot** from the header |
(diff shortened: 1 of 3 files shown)


shanehughes-ui:
Added a fifth display option in the same spirit: `charge_time_format` — the **Time to full** row reads raw minutes by default (`249 min`), or set `hm` for hours + minutes (`4h 9m`; under an hour just `45m`). Full and top-view cards. Retitled the PR to 'display options' since it's now a small family of them.

YossiKon:
Landed on `main` (`b46c8bb`), with one fix on the way in. Your own browser test `test_battery_bold_can_be_turned_off` was catching a real bug you had not seen because playwright is not in your environment: `battery_bold: false` dropped the headline to the `.num` base weight of **200** (thin), not the **400** normal your test asserts. The option names "bold on/off", so off should read normal, not thinner - the weight is now set explicitly, 700 or 400, and your test passes as written. All five options are opt-in and default to today's layout. Full suite green, coverage 100%.

PR #65, opened 2026-08-30
Fix driving state dropping out: normalise the engine-running separator

shanehughes-ui:
A car that is actually driving keeps flipping to **Parked** — on the card and in the poll loop — and a manual sync briefly shows Driving again.

**Cause:** the engine/ready state is meant to hold "driving" through the moments speed reads 0 (every light — and on the new platform the cloud snapshot's speed field simply *lags to 0* between the car's own uploads). It wasn't holding, because the value is tested against a set spelled `engine_running`, while the new platform reports it as **`engine-running`** (hyphen). The hyphen never matched, so both the card's `_isDriving` and the coordinator's `_poll_flags` fell back to **speed alone** and dropped at every 0.

Observed live on a moving car: `engineStatus` held `engine-running` for the whole trip while `speed` read `53 → 0 → 66 → 0 → 68` — and the card followed the zeros to Parked.

**Fix:** both the card and the poller normalise the separator (`-` → `_`) before the membership test, so `engine-running` and `engine_running` are equivalent. A genuinely off engine still reads parked; a trim with no engine state still falls back to speed. No other behaviour changes.

The existing tests missed this because they used the underscore spelling — both suites now also assert the real hyphenated value keeps a standstill car Driving.

Independent of the other open PRs; based on `main`.

### DIFF
--- a/custom_components/geely_connect/__init__.py
+++ b/custom_components/geely_connect/__init__.py
@@ -296,7 +296,10 @@ def _poll_flags(d: dict) -> tuple[bool, bool]:
     # stays on through those stops, so a running car counts as driving even
     # at a standstill. On a trim that never reports it this reduces to the
     # old speed test.
-    engine = str(basic.get("engineStatus", "")).strip().lower()
+    # The new platform reports "engine-running" (hyphen); the legacy set uses
+    # "engine_running". Normalise so the ready/running state matches on both -
+    # without this a moving car whose speed momentarily reads 0 flips to parked.
+    engine = str(basic.get("engineStatus", "")).strip().lower().replace("-", "_")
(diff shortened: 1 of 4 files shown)


YossiKon:
Landed on `main` (`0b2e06b`). Clean diagnosis - the value is `engine-running` on the new platform and `engine_running` on the legacy one, and the driving set only spelled the underscore, so both `_isDriving` and `_poll_flags` fell through to speed alone and dropped at every 0. Both sites normalise the separator now, `engine-off` still reads parked, and the tests pin the hyphenated value at speed 0 on the card and in the poll flags. Full suite green, coverage still 100%.

PR #66, opened 2026-08-30
Blueprint + automation: live lock-screen charging countdown

shanehughes-ui:
Adds a **live charging countdown** that stays on the phone's lock screen while the car charges:

- a **ticking countdown timer** (Android chronometer)
- the **finish time** and **charging power** on the message line
- a **charging icon pinned in the status bar** while the session runs
- it **refreshes itself silently** as the estimate changes, and is replaced by a one-line "✅ stopped — XXX km range" summary when charging ends

### What's here
Shipped both ways, matching the repo's existing pattern (raw automation + mirrored blueprint):

- `blueprints/automation/geely_connect/charging_countdown_notification.yaml` — inputs for the connection sensor, Charge Complete sensor, notify service, optional power/range sensors, status-bar icon, channel name and importance.
- a fifth entry in `automations/charging.yaml` for the copy-paste crowd.
- two README lines.

### Design notes (why it's built this way)
- **Anchored to Charge Complete, not Time To Full.** The countdown counts down to the absolute `charge_complete` timestamp, so the phone ticks it smoothly between polls — the same reasoning your existing "almost charged" automation already documents for that sensor. Time To Full is the figure from the last poll and would freeze mid-count.
- **No explicit "clear" step.** The stop-summary reuses the same `tag`, so it overwrites the pinned countdown. That avoids a race with the estimate blinking to `unknown` right at the end of a session (which `automations/charging.yaml` already warns about) — a `clear_notification` there would fight the blink and flicker.
- **`importance: default`** so it actually reaches the lock screen — Android hides low-importance notifications from it. **`alert_once`** keeps the refreshes silent after the first alert. The channel input carries a note that Android freezes a channel's importance once created, so changing importance needs a fresh channel name.
- **Android companion app** for the full effect (timer, pin, `alert_once`); iOS still gets a notification that refreshes and clears, just without the live timer.

### Testing
`tests/test_shipped_yaml.py` passes — the blueprint loads under HA's `BLUEPRINT_SCHEMA`, the automation is structurally sound, and `notify` isn't a car-command domain so the back-to-back-command guard is unaffected. Also verified the blueprint substitutes to a valid automation with real inputs (triggers, notify target, and all chronometer/channel/icon keys resolve; `when` stays a template for HA to render live).

Running live on my own EX2 (new EM/Zeekr platform) — screenshot: pinned status-bar icon + a lock-screen countdown reading "🔌 Charging · 3:32:54 / Ready by 12:21 PM · 6.8 kW".

### DIFF
--- a/README.md
+++ b/README.md
@@ -639,9 +639,9 @@ automation:
 
 Other smart behaviour: **long-term statistics** (battery, range, consumption,
 pressures feed HA statistics + the Energy dashboard) and **ready-made
-Blueprints** in `blueprints/` (charging complete, low battery, door/trunk left
-open, tire pressure out of range, left unlocked away from home, pre-condition
-climate before departure).
+Blueprints** in `blueprints/` (charging complete, a live lock-screen charging
+countdown, low battery, door/trunk left open, tire pressure out of range, left
+unlocked away from home, pre-condition climate before departure).
 
 ---
 
@@ -662,7 +662,9 @@ one entry at a time through the automation editor's ⋮ → **Edit in YAML**.
 Replace `notify.mobile_app_your_phone` with your own notifier.
 
 - **`automations/charging.yaml`** - charge finished, charge nearly finished
-  (uses the Charge Complete timestamp), low battery, and "you forgot to plug in".
+  (uses the Charge Complete timestamp), a live lock-screen charging countdown
+  (ongoing notification with a ticking timer, Android companion app), low
+  battery, and "you forgot to plug in".
 - **`automations/security.yaml`** - something left open, left unlocked after
   leaving, windows open with rain forecast, and a bedtime check.
 - **`automations/comfort-and-maintenance.yaml`** - weekday pre-heat, hot-cabin
--- a/automations/charging.yaml
+++ b/automations/charging.yaml
@@ -206,3 +206,84 @@
           ({{ states('sensor.my_geely_ex5_electric_range') }} km) and the
           cable is not connected.
   mode: single
+
+- id: geely_charging_countdown_notification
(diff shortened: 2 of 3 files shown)


shanehughes-ui:
Added a **tap-to-open** option: a `clickAction` on the notification. Blueprint gets an optional *"Open on tap (dashboard path)"* input (blank = just open the app); the raw automation uses `/lovelace/car`. Pointed at a view that shows the full `custom:geely-card`, tapping the countdown lands straight on the full card — Android can only deep-link to a Lovelace path, not fire a card's own tap-to-expand popup, so a full-card view is the way to land "full". Verified live on my EX2. Still green on `tests/test_shipped_yaml.py`.

shanehughes-ui:
Reworked completion to fire at a **true full**, not merely when charging stops: a `numeric_state` trigger on a new `battery_sensor` input (held 20s). An early unplug no longer reads as "charged" — it quietly replaces the pinned countdown with a "stopped at N%" line (no sound). The "charged" alert also moved to its **own channel** (new `complete_channel` input, default *Car charging done*, high importance) so users can give the finish a distinct sound in Android without the countdown ever making it. New inputs: `battery_sensor`, `complete_channel`, `full_level` (default 99.5). Same three-branch design mirrored in `automations/charging.yaml`. Running live on my EX2; `tests/test_shipped_yaml.py` still green.

YossiKon:
Landed on `main`. Notification-only - I checked that no vehicle-control service is called anywhere in the blueprint, only `notify` with a `clickAction` to the dashboard - and the shipped-yaml test that validates every blueprint passes. The true-full completion and the tap-to-open are both nice. Thank you.

PR #67, opened 2026-08-31
Add Thai (th) translation

tomzt:
Adds `custom_components/geely_connect/translations/th.json`, covering the config flow (platform picker, account setup, email-code, vehicle pick, the new Geely EM login step) and the options flow.

Key set matches `en.json` exactly (checked with a script, not by eye) and follows the same convention as the existing `ar`/`fr`/`he`/`ru` files.

**Scope note:** this only covers the strings routed through `strings.json`/`translations/*.json` — i.e. the setup and options dialogs. It does **not** localize the `geely-card.js` dashboard card, which has no `hass.localize()` calls today (confirmed none of the existing translated languages cover it either). Opening a separate issue to discuss adding an i18n hook there, since that's a bigger design question than a translation-file PR.


YossiKon:
Merged - thank you, @tomzt. Verified the structure before landing: valid JSON, 68 keys, an exact match to `en.json` with nothing missing and nothing extra, so the setup and options dialogs now offer Thai. I cannot vet the Thai wording itself, so if any string reads oddly in use, a follow-up is welcome.

You also opened #68 about the card being English-only - that is a real gap and correctly separate from this; I have replied there.

PR #69, opened 2026-09-01
Auto-derive the new-platform x-vin (verified against the gateway)

shanehughes-ui:
New-platform vehicles need an `x-vin` header (an encrypted VIN). Today it's an advanced owner-supplied option (`zeekr_enc_vin`); this derives it **automatically** so a new-platform car works out of the box, while keeping the manual override.

Builds on @scottaki's **#57**, which found the mechanism: `VIN → AES-128-CBC/PKCS7 → Base64`. The difference here is **the material is per-app-build and a wrong key produces a valid-looking but wrong header that only fails at request time** — so this never trusts a derived value blindly.

### How it works
- `derive_x_vin(vin, key, iv)` + `_X_VIN_MATERIAL` — the known `(key, iv)` pairs.
- `probe_x_vin(vin)` — derives with each pair and **verifies** each with a read-only, x-vin-addressed call (the capability catalogue). The gateway rejects a wrong x-vin (raises), so only a value it actually resolves to the vehicle is returned; the client's previous x-vin is restored either way.
- `config_flow` stores **only a verified** value. The manual `zeekr_enc_vin` option still overrides; an existing manual value is preserved on reauth; and **if no known build matches, nothing is stored** — behaviour is exactly as before (manual fallback). A future key rotation therefore degrades gracefully instead of hard-breaking.

### Why not a single hard-coded key (re: #57)
I tested #57's bundled key/IV against a live AU/SEA car and it does **not** reproduce that car's real x-vin (commented on #57). Different builds/regions use different material (Wysie issue #10: keys are per-version, ECIES-encrypted in newer builds). Hence: a **list** of known builds + **verify-before-use**, so an unknown build never silently mis-derives.

### Known builds shipped
- the pair from #57;
- **Geely Global EM (AU/SEA, `com.geely.global.em`)** — recovered and **validated against a real AU vehicle** (derives that car's exact working x-vin). Add more as builds are captured.

The AES material is public app package data (github.com/Wysie/zeekr_key_extractor, #57), not an account credential or session token; `enc_vin` is already redacted from logs and diagnostics.

### Tests
Derive vectors (fake VIN only) + AES round-trip + probe (verify / fallback-to-empty / restore-on-exit), plus two config-flow cases (derive-and-store, migration-stores). Full suite green locally; `coverage --fail-under=100` passes (`custom_components/geely_connect` TOTAL 100%).

### DIFF
--- a/custom_components/geely_connect/config_flow.py
+++ b/custom_components/geely_connect/config_flow.py
@@ -165,6 +165,9 @@ class GeelyIntlConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
         self._platform_default: str | None = None
         self._zeekr_hf_token: str | None = None
         self._zeekr_password: str | None = None
+        # The logged-in client from the login step, reused in _finish_zeekr to
+        # verify a derived x-vin against the gateway before storing it.
+        self._zeekr_client: Any = None
         # Set when this flow is a re-auth. We update the existing entry's
         # token instead of creating a new one in that case.
         self._reauth_entry: config_entries.ConfigEntry | None = None
@@ -520,6 +523,7 @@ class GeelyIntlConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
             try:
                 client = await self.hass.async_add_executor_job(
                     _zeekr_login_password, email, user_input["password"], self._country_code)
+                self._zeekr_client = client
                 self._zeekr_tokens = (client.access_token or "", client.refresh_token or "")
                 self._zeekr_hf_token = client.hf_token
                 self._zeekr_password = (user_input["password"]
@@ -612,6 +616,29 @@ class GeelyIntlConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
         enc_password = await self.hass.async_add_executor_job(
             password_encrypt, self.hass, self._zeekr_password or "")
 
+        # Auto-derive the new-platform x-vin from the VIN where the app build is
+        # known, verifying the gateway accepts it before storing. The
+        # owner-supplied `zeekr_enc_vin` option still overrides, and stays the
+        # fallback when no known build matches (nothing is stored then, so the
+        # entry behaves exactly as before this change). Mechanism per PR #57;
+        # the material is public app package data, not an account credential.
+        existing_x_vin = ""
+        if self._reauth_entry is not None:
+            existing_x_vin = (
+                self._reauth_entry.options.get(CONF_ZEEKR_ENC_VIN)
+                or self._reauth_entry.data.get(CONF_ZEEKR_ENC_VIN) or "")
+        enc_vin = existing_x_vin
+        if not enc_vin and self._zeekr_client is not None:
+            try:
+                enc_vin = await self.hass.async_add_executor_job(
+                    self._zeekr_client.probe_x_vin, vin) or ""
+            except Exception:  # noqa: BLE001 - derivation is best-effort only
+                _LOGGER.debug("x-vin auto-derivation failed; "
+                              "falling back to the manual option", exc_info=True)
+                enc_vin = ""
+            if enc_vin:
+                _LOGGER.info("derived a working new-platform x-vin for this vehicle")
+
         if self._reauth_entry is not None:
             entry = self._reauth_entry
             previous_email = entry.data.get(CONF_EMAIL) or ""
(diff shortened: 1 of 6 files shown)


YossiKon:
Reviewed carefully — this is a real step up from #57, and the shape is right: derive from a *list* of known builds, verify each against the gateway before storing, keep the owner-supplied `zeekr_enc_vin` as the authoritative override, and store nothing when no build matches. That directly answers the objection you raised on #57 (a single static key silently mis-derives). Security is clean: no new host or dependency, the verify call is read-only (the capability catalogue), the material is public app-package data, and it stores into the already-redacted `CONF_ZEEKR_ENC_VIN`; the x-vin block and the config-flow changes on the branch match the diff.

One substantive thing to pin down before the "verify" can be trusted, because the whole guard rests on it: **does the capability endpoint actually *reject* a wrong x-vin by raising?**

`probe_x_vin` accepts the first candidate for which `capabilities_new()` does **not** raise. But `_request` calls `_check_resp` with no `ok_codes`, so `_check_resp` raises only on an explicit `success:false` or an HTTP non-200 — **not** on a non-success `code` alone. So if the gateway answers a wrong x-vin with `HTTP 200` + `success` absent/true + empty `data`, `capabilities_new()` returns `[]` without raising, and the probe accepts that candidate. Because candidates are tried in order, that means even a car whose real build is the second (AU/SEA) pair would get **#57's** wrong x-vin stored — and an empty catalogue is read downstream as the permissive "all features" view, so it fails silently: exactly the mode this is meant to prevent. (The x-vin rides *inside* the snc signed canonical — `test_status_read_matches_the_captured_request` asserts that — so a wrong one is still validly signed and reaches the business layer; there's no signature-level rejection to lean on.)

The new test models the reject path (`fake_caps` raises for a wrong x-vin), so a green suite assumes the answer rather than settling it.

What would settle it — and you may already have it: when you validated against the live AU car, did you run `probe_x_vin` **end to end** and get the AU value (the 2nd pair)? If the probe returned the AU x-vin and not #57's, that *proves* the gateway raised for #57's wrong candidate, and the design holds. If you validated `derive_x_vin` in isolation (derived == captured), the negative case is still open — worth one deliberate check: set a knowingly-wrong x-vin and confirm `capabilities_new()` **raises** rather than returning `[]`. If it can return an empty 200, the accept test needs to be positive (a candidate that resolves to *this* vehicle), not merely "didn't raise".

None of this touches the other half: whether to embed extracted key material at all stays @YossiKon's call, same as #57. The code here is careful and the direction is the right one — this is about proving the verify step can't false-accept.

shanehughes-ui:
Great catch — I ran it end to end against the live AU car to settle exactly this, and then hardened it anyway.

**The negative case, on a real vehicle** (read-only capability calls, same session):

| x-vin | result |
|---|---|
| right (AU / 2nd pair) | `capabilities_new()` → **74 rows** |
| wrong (#57 / 1st pair) | **HTTP 400** `{"code":"079025","msg":"Decrypt X-VIN failed."}` → raises |
| garbage (32 zero bytes) | same **HTTP 400 079025** → raises |
| `probe_x_vin` end-to-end (tries #57 first, then AU) | returns the **AU** value |

So the gateway rejects a wrong x-vin at the **HTTP layer (400)**, and `_request` raises on any non-200 *before* `_check_resp` is even consulted — the "200 + success absent + empty data" path you flagged doesn't occur here. And the end-to-end probe returning the AU value proves #57's wrong candidate was rejected, not silently accepted.

**Hardened regardless** (ee73002), since you're right that "didn't raise" is a weaker contract than "resolved": acceptance is now **positive** — a candidate is kept only if `capabilities_new()` returns a **non-empty** catalogue. So even a hypothetical future build that answered a wrong x-vin with a non-raising empty 200 wouldn't be accepted (it'd fall through to the next build, then to the manual option). Added a test for that empty-non-raising case; suite green, coverage 100%.

Embedding the key material at all remains @YossiKon's call, as with #57.

YossiKon:
That settles it — thank you for running the negative case live rather than assuming it. HTTP 400 `079025 "Decrypt X-VIN failed"` for a wrong x-vin means `_request` raises on the non-200 *before* `_check_resp` is even reached, so the empty-200 false-accept path can't occur on this gateway; and the end-to-end probe returning the AU value (not #57's) is the clean proof that #57's candidate was rejected first, not silently kept.

The positive-acceptance hardening (ee73002 — keep a candidate only on a **non-empty** catalogue) closes the theoretical case too, and `test_probe_rejects_an_empty_catalogue_even_without_a_raise` covers it. Verified the loop on the branch: `if self.capabilities_new(): return candidate`, else fall through. The probe can't false-accept now — empirically or hypothetically.

So the mechanism is sound and defends itself. The one thing left is the one that was always yours to weigh, @YossiKon — whether to embed the extracted key material at all, same call as #57.

YossiKon:
Landed on main in c5d1bb2. Full gate green with the merge: **932 passed, 0 failed, 0 skipped, coverage 100%**.

What carried the decision was the verify-before-use design plus your live negative-case evidence: a wrong x-vin dies at the HTTP layer (400 `079025 Decrypt X-VIN failed`) and the positive-acceptance hardening means even a hypothetical non-raising empty answer can't store a bad token - a wrong or unknown build can only fall back to the manual option. Thank you for closing the false-accept question with real-car evidence rather than argument.

A migrated owner now sets up with email + password and nothing else. #57 closes as superseded, with the mechanism credit where it belongs.

PR #70, opened 2026-09-01
Cards: localize every visible label (#68)

### LINKED ISSUE #68: geely-card.js dashboard card has no i18n hook — hardcoded English strings
While preparing the Thai translation (#67), noticed `custom_components/geely_connect/geely-card.js` has no calls to `hass.localize()` (or any other localization mechanism) anywhere in the file. Every label the card renders - "Lock", "Climate", "Charging", "Odometer", "Find", "Maps", etc. - is a hardcoded English string.

This means `translations/*.json` (en/ar/fr/he/ru, soon th) only covers the config/options flow dialogs. The dashboard card itself stays English-only regardless of the user's HA language, for every locale, not just the new one.

Opening this as a separate issue rather than folding it into the translation PR, since localizing a Lit-based custom card is a different shape of change than adding a JSON file - it likely means either:
- routing card strings through `hass.localize()` where equivalent core HA strings already exist, plus
- a small strings map of its own (keyed by `hass.language`) for the entries that don't have a core equivalent (e.g. "Rapid Cooling", "Pack power").

Happy to help test a th (or other language) pass once there's a mechanism in place, but wanted to raise it for discussion on the right approach first rather than guessing at the architecture.

tomzt:
Closes #68.

## Mechanism

`hass.localize()` can't reach a Lovelace card - the frontend only fetches
the fixed set of translation categories it knows about (`config`,
`options`, `entity`, `state`, `services`, `selector`, `exceptions`,
`title`...), never a custom `card` one. So this follows the approach you
described after catching your own first suggestion: a `STR` object
bundled directly in `geely-card.js`, keyed by the frontend's own language
(`hass.locale.language`, falling back to `hass.language`, then `"en"`).

```js
_t(key, fallback, subs) {
  const dict = STR[this._lang(this._hass)] || STR.en;
  let out = dict[key];
  if (out == null) out = STR.en[key];
  if (out == null) out = fallback;
  if (subs) for (const k in subs) out = out.split(`{${k}}`).join(subs[k]);
  return out;
}
```

Three-tier fallback (`STR[lang]` → `STR.en` → the coded English string at
the call site) instead of falling straight to the raw key name - a button
never goes blank, and it never shows an internal key like `action.lock`
to an end user either. `{token}` substitution matches the placeholder
style `translations/*.json` already uses (`{email}`).

One thing your snippet didn't need to handle that I did: `_signature()`
(the change-detection hash `set hass()` gates re-renders on) didn't
include the language, so switching languages without a full page reload
left the card showing stale text - no watched entity had moved, so the
signature compared equal and the render was skipped. Language resolution
is now factored into `_lang(hass)`, shared by `_t()` and `_signature()`.

## Scope

Every visible label across all five cards extracted into `_t()` calls -
101 keys, `STR.en` and `STR.th` populated (this is the mechanical half;
other languages can follow the same `STR` entries once the mechanism
itself looks right to you). Covers Lock, Unlock, Climate, Charging,
Odometer, Find, Maps, the status words, the tooltips, the seat/wheel
level words (Off/Low/Medium/High - display only, doesn't touch the value
driving `_onAction`'s cycling logic), and the top-view HOOD/SUNROOF/BOOT
text.

Deliberately left alone, for reasons I'd rather you weigh in on than
guess past:
- Entity state values whose value space isn't a small closed set the card
  defines (e.g. `charger_connection`'s "Disconnected"/"Charging"/"Plugged
  in", reported by the cloud API) - translating those safely is an
  integration-level job through the `state` category HA actually loads,
  not string-matching in the card.
- Nav app brand names (Waze/Apple Maps/HERE) - "Maps" itself did get a
  key, since the button just says "Maps", not "Google Maps".
- `window.customCards` picker descriptions - pushed at module load,
  before any card instance has a `hass` to read a language from.
- Tire FL/FR/RL/RR and unit symbols (km, kW, PSI, d) - left as the
  international abbreviations HA itself doesn't translate.

## Verification

- `node --check` passes; `STR.en`/`STR.th` carry identical 101-key sets;
  every `_t()` call site's key exists in `STR.en` with none orphaned
  either direction.
- Security-reviewed (two independent passes - identification, then
  false-positive filtering on the one candidate that surfaced, a
  prototype-lookup pattern on `STR[lang]` that turned out to be
  unreachable since every key is a literal dotted string). No `_t()`-
  derived value reaches `hass.callService`, a `data-act`, or a URL - this
  is display text only.
- Built a standalone test harness (fake `hass`/entities/states, no real
  HA) exercising all five cards across en/th/an unconfigured language
  (falls back to en)/`car_image` mode - this is what caught the
  `_signature()` gap above.
- Live-tested on a real Home Assistant instance (Core, systemd install):
  restarted clean, no `geely`-related errors in the log, all five cards
  render correctly in both English and Thai, language switch without a
  page reload now updates the card.

### DIFF
--- a/custom_components/geely_connect/geely-card.js
+++ b/custom_components/geely_connect/geely-card.js
@@ -158,6 +158,207 @@
   const SWHEEL_HINT = "Steering wheel heating on / off. Confirmed on a real "
     + "car; the reading is slow, so the button holds its state for a moment.";
 
+  /* Card UI strings, bundled here because a custom `card` translation
+   * category is never fetched into the Home Assistant frontend store -
+   * only the fixed set it knows about (config, options, entity, state,
+   * services, selector, exceptions, title...) - so `hass.localize()` would
+   * silently return "" for every key here. Lovelace cards that ship their
+   * own translations (mushroom, mini-graph-card, ...) all bundle strings
+   * this way instead. `_t()` below picks the dictionary by the frontend's
+   * own language, falling back to English, then to the coded fallback. */
+  const STR = {
+    en: {
+      "action.charging": "Charging",
+      "action.climate": "Climate",
+      "action.close": "Close",
+      "action.confirm_tap": "sure?",
+      "action.defrost": "Defrost",
+      "action.find": "Find",
+      "action.lock": "Lock",
+      "action.open": "Open",
+      "action.rapid_cool": "Rapid cool",
+      "action.rapid_heat": "Rapid heat",
+      "action.schedule": "Schedule",
+      "action.sync": "Sync",
+      "action.unlock": "Unlock",
+      "action.vent": "Vent",
+      "boot.boot": "Boot",
+      "boot.trunk": "Trunk",
+      "chip.climate_on": "Climate on",
+      "chip.driving": "Driving",
+      "chip.locked": "Locked",
+      "chip.n_open": "{count} open",
+      "chip.parked": "Parked",
+      "chip.plugged_in": "Plugged in",
+      "chip.unlocked": "Unlocked",
+      "label.battery_fuel_pct": "{batt}% · {fuel}% fuel",
+      "label.driver": "Driver",
+      "label.electric_range": "Electric range",
+      "label.end": "End",
+      "label.ev_short": "{n} EV",
+      "label.fresh_air": "Fresh air",
+      "label.fuel_short": "{n} fuel",
+      "label.hood": "Hood",
+      "label.parking_comfort": "Parking comfort",
+      "label.passenger": "Passenger",
+      "label.range": "Range",
+      "label.seat_cooling": "Seat cooling",
+      "label.seat_heat": "Seat heat",
+      "label.shade": "Shade",
+      "label.start": "Start",
+      "label.sunroof": "Sunroof",
+      "label.synced": "Synced {time}",
+      "label.wheel_heat": "Wheel heat",
+      "row.battery_12v": "12 V battery",
+      "row.charger": "Charger",
+      "row.combined_range": "Combined range",
+      "row.complete_at": "Complete at",
+      "row.efficiency": "Efficiency",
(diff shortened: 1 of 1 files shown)


YossiKon:
Reviewed and verified end to end — this is well done, and it takes the corrected approach cleanly.

**Security:** only `geely-card.js`; no dependency, no network, no dynamic eval. Every `_t()` result is `esc()`'d where it enters `innerHTML`, and the `{token}` substitutions are numbers / enum labels, not free user text — so no new XSS surface.

**Mechanism:** `_lang` (locale → language → `"en"`, region stripped), `_t` with the three-tier fallback (`STR[lang]` → `STR.en` → the coded English at the call site), and folding the language into `_signature` so a language switch actually re-renders. That last one is a real catch — the stale-text bug it prevents would hit anyone changing HA language without a reload.

**No English regression:** every key `_t()` references exists in `STR.en` (91/91), and `STR.en` matches the strings the card shipped with. I ran the full suite with your card swapped in — **898 passed, 0 failed, 0 skipped**, including the Chromium card tests, so the rendered English output is unchanged.

One thing before it lands: a regression test for the mechanism itself — a `th` locale rendering a Thai label, an unknown locale falling back to English, and a missing key falling through to the coded default — since the existing browser tests only exercise the (unchanged) English. I can add that when landing, or you can push it to the branch.

Nice work for a first PR.

YossiKon:
Landed on main in f717b4d, with a browser regression test added for the mechanism — a Thai locale localizes the Lock button, `th-TH` resolves the same as `th`, and both an unknown language and a missing key fall back to English. Full suite green with the change: **900 passed, 0 failed, 0 skipped**, including the Chromium card tests.

Thank you — a clean first contribution, and folding the language into `_signature` was a sharp catch. Closing this and #68; further languages are just per-key `STR` entries from here, so they can land as small follow-ups.

PR #73, opened 2026-09-03
Cards: Windows open/close control, open indicator, and open %

shanehughes-ui:
The RWS **windows** command shipped in #60 and `cover.py` exposes a `cover.<car>_all_windows` entity, but the cards had no way to drive it — and no indication of whether the windows are open, or how far. This adds all three.

### What it does
- **Full + top cards:** a **Windows** row with **Open / Close** buttons that **lights while the windows are open** — the same `cover()` helper that already renders sunroof and sunshade, driving the existing `cover.<car>_all_windows` entity. No new command path.
- **Open %:** when a window is partly down the row shows the percentage — "Windows 8%" distinguishes a vent crack from fully open. Sunroof/shade report no position and are unchanged.
- **Compact card:** a Windows **chip** that appears *only while the windows are open*, its lit fraction tracking how far (a vent crack lights a sliver); tapping it opens a small **Open / Close** menu. The menu routes through `_onAction` (already blocked while driving), so on a moving car the chip is a read-only indicator and only acts when parked.

### Changes
**`geely-card.js`** — a `window` icon; `all_windows` added to the watched-suffix list so the indicator updates live; `windows_open`/`windows_close`/`windows_menu` actions; the cover row's windows pair with the % readout; the compact chip + its Open/Close menu; `label.windows` in the **en** and **th** dictionaries (th: หน้าต่าง — happy to have a native speaker confirm).

**`helpers.py`** — `windows_position(data)`: the most-open corner's `winPos` (0–100), clamped, junk/missing tolerated, `None` when the car omits it entirely.

**`cover.py`** — `GeelyWindows.current_cover_position` returns that, so the entity carries the position and the card can render it.

### On `winPos` (raised in review)
Confirmed present on the **new EM/Zeekr platform**. From a live `climateStatus` on a real EX2 (windows shut):

```
winPosDriver = '0'    winStatusDriver        = '2'  (2 = closed)
winPosDriverRear = '0'    winStatusDriverRear    = '2'
winPosPassenger = '0'    winStatusPassenger     = '2'
winPosPassengerRear = '0'    winStatusPassengerRear = '2'
```

The fields exist and read `'0'` because the windows are closed; a vent crack reads ~8. The legacy-platform payload on #24 has `winStatus*` but no `winPos*`, so this is a new-platform field — and `windows_position` returns `None` (no percentage shown) when it's absent, so legacy cars are unaffected. The `0–100 step 4` in my earlier note was the *command* range; the status field above is the independent evidence.

### Tests
Browser tests: full-card control (renders Open/Close, lights when open, not lit when closed, absent without the entity, press calls the cover service), the % shown/hidden, and the compact chip (present-with-fill, absent, opens-menu-and-acts). Platform test: `GeelyWindows` reports the most-open corner as its position. Full suite green, `coverage --fail-under=100` passes.

### Note
Rebased onto current `main` (resolves the earlier conflict — an additive clash in `test_cards_in_a_browser.py` with the new language-coverage test; both kept). Description refreshed to match the full diff (the open-% commits `76fe0b8`/`30ea9bb` added the `cover.py`/`helpers.py` changes after the PR was first opened).

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ dashboard, **Add card**, and search "Geely":
 
 | Card | What it is |
 |---|---|
-| **Geely Card** (`custom:geely-card`) | The full cockpit: range and battery up top, the EX5 silhouette that glows while charging and flags every opening, one-tap Lock / Unlock / Climate / Defrost / Vent / Trunk / Find / Sync, a full climate panel (temperature, rapid heat / cool, seat heating and cooling, fresh air, sunroof and shade), charging and schedule toggles with editable start / end times, links that navigate you to wherever the car is parked, then charging, tires, trip and service health - each block hides itself when the car doesn't report it |
+| **Geely Card** (`custom:geely-card`) | The full cockpit: range and battery up top, the EX5 silhouette that glows while charging and flags every opening, one-tap Lock / Unlock / Climate / Defrost / Vent / Trunk / Find / Sync, a full climate panel (temperature, rapid heat / cool, seat heating and cooling, fresh air, windows, sunroof and shade), charging and schedule toggles with editable start / end times, links that navigate you to wherever the car is parked, then charging, tires, trip and service health - each block hides itself when the car doesn't report it |
 | **Geely Card (top view)** (`custom:geely-card-top`) | The car from above: tire pressure beside each wheel, bold live status on every door, the hood, the sunroof and the trunk - with the same header, actions and climate panel. The driver's door follows your market automatically (right-hand-drive countries show it on the right; `rhd: true/false` in the card config overrides) |
 | **Geely Card (compact)** (`custom:geely-card-compact`) | The essentials in one tile: battery and cabin temperature in the header, range, status chips - and Lock, Unlock, rapid Heat / Cool, Defrost and Trunk |
 | **Geely Card (mini)** (`custom:geely-card-mini`) | A small square: range, battery and cabin temperature in the header, status, a lock button that follows the car (locked offers Unlock, unlocked offers Lock), and one-tap quick heat / quick cool |
@@ -342,8 +342,9 @@ The full and top-view cards carry a complete **Climate** section:
 - **Seat heating / Seat cooling** per front seat - each tap steps
   Off → Low → Medium → High → Off.
 - **Fresh air** (G-Clean), **Parking comfort**, **Wheel heat** (on cars that
-  have a heated steering wheel), and **Open / Close** for the sunroof and the
-  sunshade. Parking comfort never lights up, deliberately: the car does not
+  have a heated steering wheel), and **Open / Close** for the windows, the
+  sunroof and the sunshade - each pair lights while that opening is open.
+  Parking comfort never lights up, deliberately: the car does not
   report whether it is on - the field that looks like its flag reads 1 with the
   feature off - so the button toggles it without claiming a state.
 - Every block hides itself on a trim that lacks the entity - and there is no
--- a/custom_components/geely_connect/cover.py
+++ b/custom_components/geely_connect/cover.py
@@ -43,7 +43,7 @@ from homeassistant.helpers.update_coordinator import CoordinatorEntity
 
 from .api import GeelyControlError, redact
 from .const import DOMAIN, SERVICE_WINDOW
-from .helpers import walk as _walk, windows_open, schedule_refresh
+from .helpers import walk as _walk, windows_open, windows_position, schedule_refresh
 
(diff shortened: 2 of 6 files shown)


shanehughes-ui:
Extended this with the window **open %**, since the car reports a per-corner `winPos` (0-100; a vent crack reads ~8). `GeelyWindows` now exposes `current_cover_position` (most-open corner, via a new `helpers.windows_position`), and the card's cover row shows the % when a cover is partly open — so "Windows 8%" distinguishes a vent crack from fully down. Sunroof/shade report no position and are unchanged. Tests + browser tests added; full suite green, coverage 100%. Retitled accordingly.

shanehughes-ui:
Added the same to the **compact card** as a chip: it appears only while the windows are open, its lit fraction tracks how far (a vent crack ~8% lights a sliver), it reads e.g. "Windows 8%", and tapping it opens a small **Open / Close** menu. The menu routes through `_onAction` (already blocked while driving), so on a moving car the chip is a read-only indicator and only acts when parked. Browser tests cover the chip's presence/fill and the tap-menu firing the cover service. Full suite green (942 passed), coverage 100%.

YossiKon:
The card half is good and I'd take it as it stands - the Windows row reuses the same `cover()` helper as sunroof and sunshade, drives the `cover.<car>_all_windows` entity that already exists, and the four browser tests cover render / lit / not-lit / absent. No new command path, no new plumbing.

Two things before it lands, both about the half that arrived after the description was written (`76fe0b8` and `30ea9bb`, which add the open **%**):

**1. `winPos*` needs a capture.** The name appears nowhere else in the repo, and it is absent from the one real payload I have in front of me - a legacy-platform EX2 diagnostics posted on #24 yesterday, whose `climateStatus` carries `winStatusDriver`/`winStatusPassenger`/… and their `*Validity` partners, but no `winPos` anything. Your note says the command range is 0-100 step 4, which is the *command* parameter - that is not evidence of a status field by the same name. If you have a payload showing `winPosDriver` on a car, post it and this goes straight in; the tests can't settle it because they build the field themselves. The failure mode is benign (`None` → no percentage), so this is about not shipping a claim we can't show.

**2. The description is stale.** It says "Changes (all in `geely-card.js`)" and "no Python changed", but the branch also changes `cover.py` and `helpers.py` - the PR was opened right after the first commit and the later two are not in it. Worth a refresh so the record matches the diff.

Everything about the Python itself is careful, for what it's worth: clamped 0-100, junk and missing fields tolerated, `None` rather than a confident 0, and tested. It is only the field's existence I can't confirm.

shanehughes-ui:
Both points addressed.

**1. `winPos` is confirmed — it's a new-EM-platform field.** Pulled a live `climateStatus` off a real EX2 on the new platform (via the integration's own diagnostics), windows shut:

```
winPosDriver        = '0'    winStatusDriver        = '2'   (2 = closed)
winPosDriverRear    = '0'    winStatusDriverRear    = '2'
winPosPassenger     = '0'    winStatusPassenger     = '2'
winPosPassengerRear = '0'    winStatusPassengerRear = '2'
```

The `winPos*` fields are present and read `'0'` because the windows are closed (their `winStatus*` partners read `'2'`); a vent crack reads ~8. That's the field itself, not the command range — my earlier "0-100 step 4" note was the command parameter, which you rightly said proves nothing. The #24 payload is a *legacy*-platform car, which is why it has `winStatus*` but no `winPos*`. `windows_position` returns `None` when the field is absent, so legacy cars simply show no percentage — the benign failure mode you noted.

**2. Description refreshed** to match the full diff — it now covers `cover.py`/`helpers.py` and the open-% (the `76fe0b8`/`30ea9bb` work that arrived after the PR was opened), plus the compact chip.

Also **rebased onto current `main`** to clear the conflict. The only clash was additive — your new `test_every_shipped_language_covers_the_english_key_set` landed at the same spot in `test_cards_in_a_browser.py`; kept both. That test then correctly caught that my new `label.windows` key needed an Italian entry (added `Finestrini`). Full suite green, coverage 100%.

YossiKon:
Landed in 583dfd2. The `winPos` capture is what it needed - `winPos*` at 0 beside `winStatus*` at 2 on shut windows is the field itself, and it also explains why the legacy payload on #24 has the status fields and not the position ones. Refreshed description matches the diff now too.

Nice symmetry: the translation guard from 79695fb caught your new `label.windows` needing an Italian entry, which is exactly the gap it was written for. Re-verified here after the rebase - 951 tests, coverage 100%.

PR #74, opened 2026-09-04
Charging notification: reminder for "plugged in but not charging"

shanehughes-ui:
## What

Adds a fourth branch to the charging-countdown notification: a **"plugged in but not charging"** reminder, for the case where the car is plugged in but drawing no power — a wall switch left off at the socket, a scheduled charge that never armed, etc. Today the countdown just stays silent in that state, and the owner only finds out hours later that the car never charged.

> 🔌 **Plugged in — not charging** · Check the wall switch — the car isn't charging (battery 47%).

## How

- A delayed trigger `to: "Plugged in"` with `for:` (grace period) and `id: plug_grace`. The `id` means the reminder branch fires **only off this timer**, never the instant the plug goes in — so a normal plug-in gets time to start charging (and the cloud gets time to report it) before anything nags.
- The branch fires only while **still idle** (`connection == "Plugged in"`) and **below a reminder level** (default 95%). Above that, being plugged in and idle is normal — the car's as good as full and the existing "charged" alert covers the top end.
- Its **own channel/tag** (`geely_plug_reminder`), so it can carry its own sound/importance and be cleared cleanly. It's retired the moment charging begins (added to the charging branch) and at a true full.

## Inputs (blueprint)

Three new inputs, all with sensible defaults so existing users are unaffected:

| Input | Default | |
|---|---|---|
| Plugged-in reminder: only below (%) | `95` | set `0` to disable the reminder entirely |
| Plugged-in reminder: wait (minutes) | `3` | grace period before it fires |
| Plugged-in reminder channel | `Car plugged in` | its own Android channel |

The paste-in `automations/charging.yaml` mirror carries the same logic as `# TUNE:`-commented literals.

## Notes / testing

- Both files parse and keep the existing structure; every added step is a `notify` call, so `test_no_shipped_yaml_fires_two_car_commands_back_to_back` stays green and no remote command is involved.
- Verified against a real session on my own EX2: on this platform the connection sensor reports `Plugged in` (idle) vs `Charging`, and the reminder correctly distinguishes the two. Running live on my box now.

### DIFF
--- a/automations/charging.yaml
+++ b/automations/charging.yaml
@@ -231,6 +231,16 @@
       entity_id: sensor.my_geely_ex5_battery
       above: 99.5
       for: "00:00:20"
+    # Plugged in but still not charging after a grace period -> wall-switch
+    # reminder (a switch left off at the socket). The "for" gives a normal
+    # plug-in time to start - and the cloud time to report it - and the id lets
+    # the reminder branch fire only off this timer, never the instant you plug
+    # in. TUNE: 00:03:00 grace.
+    - trigger: state
+      entity_id: sensor.my_geely_ex5_charger_connection
+      to: "Plugged in"
+      for: "00:03:00"
+      id: plug_grace
   actions:
     - choose:
         # TRUE full: clear the countdown, then a "charged" alert on its OWN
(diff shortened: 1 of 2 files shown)


YossiKon:
Landed in 2dc6c69. The design is the careful part: it fires only off its own delayed trigger rather than on the plug-in edge, it stays quiet near full where plugged-and-idle is normal, and it retires its own tag both when charging starts and when the car reaches full - so it cannot leave a stale reminder pinned. New inputs all have defaults, and 0 turns it off, so an existing automation is unchanged.

Shipped-YAML checks and the full gate green here (955 tests, coverage 100%).

PR #75, opened 2026-09-04
Add Italian translation to geely-card.js

manuele84:
Aggiunge traduzione italiana completa, 100 chiavi, stessa struttura di th


YossiKon:
Landed, with one string changed - thank you, and please correct me if my Italian is off.

**The dictionary checks out.** 101 keys, exactly the English set: nothing missing, nothing extra, and every `{token}` preserved. Purely additive to `geely-card.js`.

**The one change.** `status.n_openings_open` came through as `"{count} apertur{plural} aperte"`, and `{plural}` is not a plural-aware token - the card substitutes the English letter `"s"` (or `""`):

```
1 opening  -> "1 apertur aperte"
2 openings -> "2 aperturs aperte"
```

Both wrong, and my fault for not documenting the token when the mechanism shipped. I reworded it to phrase around the count instead:

```
"status.n_openings_open": "Aperture aperte: {count}"
```

which reads correctly at 1 and at N. If a native speaker would put it differently, say so and I'll take your wording.

**Guarded for the next language.** Added a test that reads every shipped dictionary and fails on: a missing or unknown key, an empty value, an invented `{token}` the card never substitutes, and any use of `{plural}` outside English. It also confirmed Thai legitimately *drops* `{plural}` (Thai has no grammatical plural) - dropping a token is fine, inventing one is not.

Full browser suite green with Italian in: 98 passed, 0 failed, 0 skipped.

PR #76, opened 2026-09-06
Compact card: opt-in charging countdown (ready-by + time left)

shanehughes-ui:
Adds an opt-in compact-card option that reworks how a charging session reads, for owners who want the finish time and remaining time at a glance rather than just the kW.

### What it does (when enabled)
- The live **power moves up to a status line under the title** — `Charging · 1.7 kW` — the same treatment the full card already gives it.
- The charging **chip becomes a countdown**: `Ready 10:22 · 2h 15m left`, derived from `charge_complete`.
- It **ticks down every minute** so the "time left" stays current between polls (the car only reports new data every few minutes).

### Opt-in — default is unchanged
New compact-card config option **`charging_countdown:` (default `false`)**. Unset, the card behaves exactly as today (the `⚡ 1.7 kW` chip, no status line), so existing users see no difference. `charge_complete` is only added to the watched set when the option is on, so a default card keeps its render signature.

```yaml
type: custom:geely-card-compact
prefix: my_geely_ex5
charging_countdown: true
```

### Implementation notes
- The minute tick is an idempotent timer on `GeelyCardBase` (`_ensureMinuteTick`): a render arms it only while a countdown is on screen and clears it the moment charging stops; `disconnectedCallback` clears it too, so a removed or parked card never keeps re-rendering.
- Countdown from `charge_complete` (absolute timestamp), formatted with the card's existing `toLocaleTimeString` style; the `h m` remaining matches the existing "Time to full" `hm` format. Falls back to a plain `Charging` chip (and no tick) when the car hasn't reported a finish time yet.
- New translation key `chip.charge_ready` in **en/th/it**; the status line reuses the existing `status.charging_line`.

### Tests
Four browser tests: opt-in shows the status line + ready/left chip and arms the tick; default keeps the kW chip and no status line; no-finish-time falls back to `Charging` with no tick; the tick is cleared when charging ends. Full browser suite green (102), language-coverage test green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,7 @@ Everything else is optional:
 | `battery_bold:` | `true` | Set `false` to draw the battery-% headline in a regular weight instead of bold |
 | `charge_time_format:` | `min` | How **Time to full** reads while charging: `min` (e.g. `249 min`) or `hm` for hours + minutes (e.g. `4h 9m`) |
 | `show_parked:` | `false` | Compact card only. `true` adds a **Parked** chip next to Locked while the car is stationary, matching the full card's status |
+| `charging_countdown:` | `false` | Compact card only. `true` moves the live charging power up to a **status line under the title** (as the full card shows it) and turns the charging chip into a **ready-by time + countdown** (e.g. `Ready 10:22 · 2h 15m left`) that ticks down each minute |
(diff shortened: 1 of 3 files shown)


YossiKon:
Landed in a05dc23. The part I looked hardest at was the minute timer, and it is right: created once, only while a countdown is actually on screen, and cleared in `disconnectedCallback` - so navigating off a dashboard leaves nothing ticking. Default-off with `charge_complete` outside the watched set until enabled means an existing card is untouched.

One thing worth recording: its `geely-card.js` hunk conflicted with #73's windows row, which landed minutes earlier. Both sides were additive so I kept both, then proved it rather than assumed it - the 111 browser tests exercise the windows control and the countdown together, and the full gate is 955 at 100%.

Issue #17, opened 2026-08-05
Wrong voltage x current in charging


alexandrenakagawa:
6kW wallbox, 240V x 32A. The integration shows strange values like 1,500V and 16A. It seems this might actually be 1.5V from some cell voltage.

<img width="563" height="481" alt="Image" src="https://github.com/user-attachments/assets/905271f8-0aa3-4641-bb94-a808f7e38b59" />

<img width="1789" height="493" alt="Image" src="https://github.com/user-attachments/assets/82d0979a-e288-4d45-94c4-eadf0f3ff580" />

YossiKon:
Your instinct is right that the number is mis-scaled, and the product math confirms the chain: 1581.0 x 16.3 / 1000 = exactly the 25.77 kW the card showed - so the power sensor faithfully multiplied a garbage voltage. What that field actually encodes on your car (cell millivolts? a tenth-scale error?) is still unknown, but it does not need to be known to stop the damage:

**v1.21.1** puts a plausibility wall in front of the AC pair - no real EVSE exceeds ~500 V or ~150 A, so readings beyond that lose the leg election, and the pack pair (the DC side, ~460 V at about -13 A during your session) carries the reading instead: roughly **6.3 kW**, which matches your wallbox.

Two things that would finish this properly, when convenient:
1. A **diagnostics download** taken *while AC charging* (Settings - Devices & services - Geely Connect - your car - Download diagnostics) - it shows the raw `chargeUAct` / `chargeIAct`, which would tell us what the 1581 actually is. If it turns out to be a clean factor (say, deci-volts of something real), the value could be rescaled rather than rejected.
2. After updating, confirm the Charging Power reads ~6 kW on the same wallbox.

Thanks for the sharp report - two good catches from Brazil in one week.

alexandrenakagawa:
i did a debug service in HA.
I'm charging from an AC wallbox (approximately 236 V / 28.4 A ≈ 6.7 kW). The raw API returns:

chargeUAct = 236.7
chargeIAct = 28.400
dcChargeUAct = 1586
dcChargeIAct = -16.1

The integration currently reports a charging voltage of 1586 V, but the raw payload suggests that chargeUAct is the correct AC charging voltage while dcChargeUAct appears to be unrelated (or invalid during AC charging). Could the integration be using dcChargeUAct instead of chargeUAct for AC sessions?

        "electricVehicleStatus": {
          "chargeLevel": "72",
          "statusOfChargerConnection": "3",
[...]

PR #79, opened 2026-09-08
Add DC-only charge current and voltage sensors

shanehughes-ui:
Two diagnostic sensors that surface a DC fast charge on its own: **DC Charge Current** and **DC Charge Voltage**.

### Why
The existing **Charge Current / Voltage** entities already switch to the DC pair via `_charge_leg` during a DC session, but the voltage they show is the *pack* (`dcChargeUAct`) — the field a car can send mis-scaled (1586 V, #17). The car also reports `dcChargePileUAct`, the **charger's own output voltage** (e.g. 396 V), which nothing surfaced. These two entities give a clean DC-only view — blank on AC and idle — so a dashboard can show a fast charge without the AC/DC entities' idle noise.

### What
- **DC Charge Current** ← magnitude of `dcChargeIAct`, read off the elected leg.
- **DC Charge Voltage** ← `dcChargePileUAct` (the pile/charger voltage), with the same pack-voltage plausibility window Pack Power uses as a backstop.

### The care that went in
- **Gated on an active DC session** via a shared `_dc_charging(data)` helper that reuses the vetted `_charge_leg` election — so the sensors light up on exactly the sessions Charging Power calls DC. This matters because the DC fields read **stale** when idle: a parked (or *driving*) car still reports a `dcChargePileUAct` from its last fast charge. A gap beats a stale reading.
- The current reads its magnitude off the leg (which is only DC once `dcChargeIAct` has parsed), so there's no second failure mode to guard. The voltage rejects an absent or out-of-window pile reading as `None`.
- Behind the same propulsion `charges` gate as the other charge entities, and added to the charging-entity gate test so a socketless car still gets none.

### Notes / testing
- Five tests: a live DC session (current = magnitude, voltage = pile), idle with stale fields (both blank), an AC session carrying a stale pile reading (both blank), an implausible pile voltage (voltage blank, current still reports), and an absent pile field (voltage blank).
- README entity counts updated 84/97 → 86/99 (the documented-counts test enforces these).
- `winPos`-style honesty: confirmed the fields exist and are populated on a real new-EM-platform EX2; the exact readings *during* a DC session are inferred from the integration's existing #10 DC log, since I have no DC session of my own to capture. Fails safe (`None`) if a real session's fields differ.
- Full `charge_power` and `entities` suites green; new lines are 100% covered. Running live on my own car.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ polished setup on top.
 ## ✨ Highlights
 
 - 🔒 **Security-first** - verified TLS plus public-key pinning.
-- 📊 **Everything enabled** - all 84 entities are on from the start (97 on a
+- 📊 **Everything enabled** - all 86 entities are on from the start (99 on a
   hybrid), no duplicates, nothing to switch on by hand.
 - 🧮 **Computed extras** - charging power, charge completion time, range at full
   charge and efficiency, none of which the car reports itself.
@@ -1118,7 +1118,7 @@ enabling. Everything the car reports, plus the computed extras above.
 
 The main thing that varies by car is propulsion: the thirteen fuel and engine
 entities are created only for a car with a tank, so a battery-electric EX5 gets
-84 entities and a PHEV gets 97. That's a decision made once at startup from your
+86 entities and a PHEV gets 99. That's a decision made once at startup from your
 account's `powerType` plus the car's own telemetry - there is no option to set.
 
 Two more - `Charging (reported)` and `Plugged In (reported)` - are built only for
--- a/custom_components/geely_connect/sensor.py
+++ b/custom_components/geely_connect/sensor.py
@@ -556,6 +556,11 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry, add_entitie
            GeelyChargePowerSensor(coordinator, vin, device_name),
            GeelyChargeCurrentSensor(coordinator, vin, device_name),
            GeelyChargeVoltageSensor(coordinator, vin, device_name),
+           # DC-only views of a fast charge - the pile's current and its own
+           # output voltage - blank on AC and idle, so a dashboard can show a
+           # fast charge on its own without the AC/DC entities' idle noise.
+           GeelyDCChargeCurrentSensor(coordinator, vin, device_name),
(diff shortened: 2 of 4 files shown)


YossiKon:
Landed in dc4bbd0.

The one thing I wanted settled before taking it was whether `dcChargePileUAct` actually *reaches* the sensor on the platform you saw it on. It does, and not by accident: `_wrap_vehicle_status` in the adapter only re-nests the new gateway's top-level keys - `additionalVehicleStatus` is moved whole, never rebuilt key by key - so anything the new gateway reports inside `electricVehicleStatus` arrives at `_walk` unchanged. Had the adapter mapped fields by name, this entity would have been permanently blank on exactly the cars that have the field.

Two changes on the way in, neither of them to the behaviour:

- The two gates became one `_dc_leg(data)` resolver returning the leg or `None`, for the same reason `_charge_leg` is one resolver for three sensors. Folding them also dropped the `leg[0] > 0 or leg[1] > 0` clause, which is unreachable: the DC branch of `_charge_leg` already requires `abs(dc[1]) > 0` and a voltage inside the pack window, and the not-charging path returns the AC leg.
- A README entity-table row. The counts test catches a stale number but not a missing row, and every other entity has one.

966 tests, 100% coverage locally. Worth saying plainly: CI never ran on this PR - fork runs sit at `action_required` here - so that green is mine, not Actions'.

Your caveat is carried into the commit message rather than dropped, because it is the honest state of these two entities: the fields are confirmed present on a real new-EM-platform car, the readings *during* a session are inferred from the #10 log, and both fail to a gap if a real session disagrees. So the standing ask, for anyone reading this thread: **diagnostics captured during an actual DC fast charge**. Nobody has sent one yet, and it is the only thing that turns this from inferred to measured.

One inherited edge, stated so it is on the record rather than found later: if a car ever reports `statusOfChargerConnection` 3 with an implausible AC pair and an in-window DC pair, `_charge_leg`'s contactor-open fallback elects the DC leg, and these two would light up on what is really an AC session. Pre-existing, no report of it, and not worth rewriting the election over - but it is the one case your "stays blank through AC" test cannot cover.

